PR #74, opened 2026-07-05
fix(leaderboard): reject partial LLM fallback below 95% coverage (H6)

FlyM1ss:
## Problem

The H6 integrity guard (`_reject_if_llm_fallback`) only refused **total** fallback — `used_llm` False or `llm_calls == 0`. A run where the model decided a *few* steps and fell back to rule-based for the rest passed the guard and published under the model's name. In practice a **1-of-161** run (≈99% rule-based) silently topped the leaderboard.

## Fix

Add a **coverage threshold**: an LLM entry must have decided at least `MIN_LLM_DECISION_COVERAGE` (0.95) of its steps with the model itself, or publishing is refused (unless `allow_fallback=True`). The 5% margin tolerates transient API blips on a genuine run (e.g. 159/161) while catching partial-fallback curves.

- Expose `LLMAgentStrategy.decision_steps` (total steps offered to the model) as the guard denominator. `llm_calls` already counts only *successful* LLM decisions (any exception path falls back to rule-based without incrementing), so `llm_calls / decision_steps` is a truthful coverage ratio.
- `_reject_if_llm_fallback` now enforces the threshold on **both** insert paths (`deploy_model_run` + `ensure_leaderboard_runs`). Coverage is checked only when `decision_steps` is known (>0); a genuine run always reports it.
- Rule-based baselines (no `used_llm`) and `allow_fallback=True` are unaffected.

## Tests (TDD)

New cases in `test_deploy_guard.py`, all green:
- `test_refuses_partial_llm_fallback` — the real 1/161 case → refused, nothing persisted.
- `test_refuses_just_below_coverage_threshold` — 94/100 (94%) → refused.
- `test_publishes_at_coverage_threshold` — 95/100 (exactly 95%) → published (guards against over-strictness).

Full leaderboard suite: 24 passed. The 3 pre-existing `test_deleted_shim_is_not_importable` failures on `main` are unrelated to this change.

### DIFF
--- a/dashboard/backend/domain/backtesting/portfolio_manager.py
+++ b/dashboard/backend/domain/backtesting/portfolio_manager.py
@@ -64,7 +64,8 @@ class PortfolioManager:
         self.trades = []
         self.equity_history = []
         # Real LLM token usage (server-side calls report actual counts)
-        self.llm_calls = 0
+        self.llm_calls = 0        # billed API calls (any response with usage)
+        self.llm_decisions = 0    # steps the model actually drove (H6 coverage)
         self.input_tokens = 0
         self.output_tokens = 0
     
@@ -392,9 +393,18 @@ class PortfolioManager:
                 
                 # else: HOLD is implicit (don't add to actions)
             
+            # Count this step as model-driven only here, at the single success
+            # exit — the model's actions were parsed AND processed without error,
+            # so the returned decision is genuinely the model's. This is the H6
+            # coverage numerator, distinct from llm_calls (billed calls): any
+            # exception in the processing loop above is caught below and swapped
+            # for a rule-based decision, which must NOT count. (Actions filtered
+            # for low confidence / invalid symbol still count: the model drove
+            # the step, our policy merely declined to act on it.)
+            self.llm_decisions += 1
             print(f"   ✅ Total actions: {len(actions)}\n")
             return {"actions": actions}
-        
+
         except Exception as e:
             print(f"\n❌ LLM decision error: {e}")
             print(f"   Falling back to rule-based logic\n")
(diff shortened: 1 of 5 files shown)


FlyM1ss:
## Follow-up commit (05d3974): review fixes

Addresses findings from a review of this PR (including an adversarial verification pass that reproduced a live defect in the first cut of the fix).

**Core issue found:** the coverage guard's numerator was `llm_calls` — *billed* API calls. `llm_calls` increments on any response with token usage, so a run whose model reliably emits truncated/unparseable output, or valid-JSON-but-malformed action items (e.g. `confidence` as the string `"0.9"`), would show **100% coverage while trading fully rule-based** — the exact "looks LLM-driven but isn't" shape this guard exists to catch, just via a post-parse failure mode instead of a total API failure.

**Changes:**
- **New `llm_decisions` counter** (steps the model actually drove), incremented only at the *single success exit* of `make_trading_decision_with_llm` — after the action-processing loop completes. Any exception in that loop is caught by the blanket `except` and swapped for a rule-based decision, which now correctly does **not** count. `llm_calls` stays a pure billing counter (DB/cost reporting unchanged).
- **Guard keys coverage off `llm_decisions / decision_steps`**, so `llm_calls == decision_steps` with `llm_decisions ≈ 0` is refused.
- **`model_id` now named** in the `ensure_leaderboard_runs` (auto-compute) diagnostic — it previously printed `'None'`.
- **`_reported_int` helper** so the documented "`llm_decisions` defaults to `llm_calls` for strategies that don't report it" contract is actually reachable (call sites previously coerced a missing attribute to `0`, wrongly rejecting a genuine older-shape run).

**Tests (TDD):** counter counts only usable model-driven steps (malformed / empty / post-parse-exception are billed-but-not-a-decision; filtered actions still count); guard refuses full-calls-but-no-usable-decisions; `allow_fallback` bypasses the partial guard too; missing-`llm_decisions` strategy defaults to `llm_calls`. Full backend suite **854 passed** (3 pre-existing `test_deleted_shim` failures unrelated).

PR #102, opened 2026-07-13
feat: FinSearch news & signals home panel

FlyM1ss:
Plan 1 (Phase A) consumer half of the FinSearch news→sentiment→ATL bridge. Spec/plan: `docs/superpowers/{specs,plans}/2026-07-13-finsearch-news-signals-panel*`.

**Ships**
- `integrations/news_sentiment.py` — fail-closed adapter (shared HTTP core + memo policy: past-date hard-memo incl. negative-cached config errors, today/latest always ETag-revalidate, transport failure never serves stale). Populates the v2 `news_sentiment` envelope slot; ISO-string timestamp coercion (the real call site passes a string).
- `GET /api/news/signals` — sync proxy (single-flight lock, 420s TTL cache, 30s negative-cache), always 200 fail-closed. Bearer key stays server-side.
- Home panel (`#homeView`): latest stories (left) + identified signals (right), poll-while-visible, `safeUrl`-before-`escapeHtml` on every href.
- Additive optional `rationale` on `NewsSentimentEntry` (v2 contract) + normative-doc updates.
- `.env.example` (`FINGPT_API_KEY`, `FINSEARCH_SIGNALS_URL`), README roadmap + agent-API doc.

**Tests:** full suite **927 passed / 2 skipped / 0 failed**. Adapter/route/contract tests are offline (vendored fixture, no network, no key). Both route-contract freezes updated (no CI reddening).

**Review:** each task spec+quality reviewed with fixes applied; final whole-branch review = ready to merge, 0 Critical/0 Important. All four cross-task seams (payload shape adapter→route→frontend, fail-closed chain, key-never-crosses-browser, memo/TTL interaction) verified clean.

**Manual follow-ups (not in this PR)**
- Browser visual QA of the panel (code + server-side smoke verified; headless can't drive a browser).
- Set `FINGPT_API_KEY` in Render prod env (same value as the FinSearch backend).
- Phase B / Task 8 (cross-repo FinSearch raw-items endpoint + feed upgrade) is a separate PR.
- Non-blocking test-fidelity follow-up: add a stripped-wire-shape fixture (exercises `staleness_hours`) + a `degraded`-status branch test.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,6 +13,15 @@ DEEPSEEK_API_KEY=your_deepseek_key_here
 COMMONSTACK_API_KEY=your_commonstack_key_here
 COMMONSTACK_BASE_URL=https://api.commonstack.ai
 
+# Agentic FinSearch news-signals producer (bearer key, same value as the
+# FinSearch backend). Required for the Home news panel and backtest
+# news_sentiment to populate; leave unset and the panel shows "unavailable"
+# and backtests fail closed to an empty {} slot.
+FINGPT_API_KEY=your_fingpt_api_key_here
+# Optional override of the FinSearch producer base URL (defaults to the live
+# endpoint). Point at a local fixture/staging server for dev.
+# FINSEARCH_SIGNALS_URL=http://localhost:9000
+
 # OpenRouter multi-provider gateway (Anthropic Messages skin).
 # Leaderboard entries with "integration": "openrouter" route through this
 # (e.g. nvidia/nemotron-3-nano-30b-a3b). Opt-in only — never auto-selected.
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ AgenticTrading/
 ## Future Roadmap
 
 - Leaderboard backed by real multi-agent runs (replace mock data)
-- Sentiment analysis (Reddit, news APIs)
+- News sentiment is live via Agentic FinSearch (Home panel + v2 agent context); Reddit/social sentiment still planned
 - Monte Carlo simulation baselines
 - Production-ready Docker image (frontend + data included)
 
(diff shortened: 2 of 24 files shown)


FlyM1ss:
Addressed the review findings in 73734dd (942 passed, 2 skipped):

| # | Finding | Fix |
|---|---------|-----|
| 1 | transport failure never memoized → N×10s backtest hangs | `fetch_signals` now caches the gap on transport failure (past: permanent, today: short throttle) |
| 2 | stampede + outage-throttle lived only in the router | single-flight + negative-cache moved into the shared adapter; non-blocking followers |
| 3 | `_fetch_lock` held across the ~10s call in the threadpool | new `PaperTradingCache.get_or_fetch`; router drops the global lock |
| 4 | `staleness_hours.toFixed` throws on a non-number | `Number(...)` + `Number.isFinite` guard (null still → "") |
| 5 | one malformed signal collapses the whole panel | feed projection guards each signal; bad entries dropped/logged, sort tolerates bad `published` |
| 6 | `_coerce_timestamp` missing the `"Z"` fix | normalizes trailing `Z` like `equity_plot.parse_equity_timestamp` |
| 7 | duplicated story-field projection | shared `_story_fields` helper |
| 8 | doc omits the 404/no-artifact case | rst now names it (the common historical-backtest case) |

Lower-priority: added a fixture content-parity guard; the `get_or_fetch` primitive replaces the hand-rolled double-checked locking.

Note on #2/#3: followers are **non-blocking** — during a genuine cold-start fetch a concurrent caller gets a brief `unavailable` (self-heals next poll) rather than holding a threadpool worker for ~10s. Tradeoff chosen deliberately to protect the deadline-sensitive `/api/v1/runs/*` endpoints that share the pool.

PR #107, opened 2026-07-14
feat(news): panel feed prefers raw items endpoint

FlyM1ss:
Phase B of the FinSearch news-signals integration (Phase A = #102). Upgrades the Home panel's **feed** (left column) to prefer FinSearch's new `GET /api/news/items/` raw-items endpoint, falling back to the existing Phase-A representative-story feed on **any** failure — so the panel can never regress below Phase A.

**What it does**
- `get_latest_panel_payload` now fetches `/api/news/items/` first for the feed; on any failure (transport error, non-200, unparseable/non-dict body, empty, all-malformed) it falls back to the signals-derived representative feed. `signals`/`status`/`overview`/`staleness` still come from the signals artifact (unchanged).
- Panel feed output shape is **unchanged** (`{headline, source, url, published, ticker}`); the AF wire now speaks news-story v1 (`headline`/`url` + `schema_version`). Frontend: heading → "Latest news", null-ticker meta-line fix.
- New `FINSEARCH_ITEMS_URL` env (`.env.example` updated); `fetch_items` is fully fail-closed (never raises).

**Tests**: +9 offline cases in `test_news_sentiment_adapter.py` (mapping correctness, each fallback trigger, signals-down-skips-items-fetch, non-dict body). Adapter file 30/30.

**Depends on**: the FinSearch `/api/news/items/` endpoint being deployed (separate PR in Agentic-FinSearch). Until then the feed transparently uses the Phase-A source.

**Note on CI**: the one failing test (`test_agent_runs_metadata.py::test_engine_llm_run_metadata_snapshot`) is **pre-existing on `main`**, unrelated to this PR — introduced by #104 (`_llm_run_metadata` reads `self.prompt_adaptations`, which the `__new__`-based test never sets). Tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018rCr1NSRDG8kpsYtHPndmM

PR #104, opened 2026-07-14
post-trade analysis

JinBoatus1:
## Summary
- Add a default **Post-trade Analysis** sub-agent that runs **once per trading day** (not hourly), after trades settle.
- Uses that day's equity change, trades, latest step outputs, and current prompts as feedback to propose and apply `prompt_patches` to upstream decision steps for the next day.
- Persist adaptations in run `metadata`; when a backtest finishes with an `agent_id`, write the adapted pipeline back to the agent.

PR #110, opened 2026-07-14
fix(news): read news-story v1 wire shape (headline/url)

FlyM1ss:
AF #359 renamed the `/api/news/items/` wire fields to `headline`/`url` (news-story v1) and went live in prod at 13:14 UTC today. This adapter still reads `title`/`link`, so every item hits the drop-malformed path, the feed empties, and `if not feed:` silently falls back to the Phase-A representative stories. No UI error — only `dropping malformed panel item` warnings in the log.

A/B against live prod, `_feed_from_items()` on 5 real items:

| adapter | raw items | feed entries |
|---|---|---|
| `main` | 5 | 0 — Phase-A fallback |
| this branch | 5 | 5 — real items |

Also carries the "Latest news" heading + null-ticker meta line (review finding #1) and the news-story v1 contract doc.

Tests: full backend suite 957 passed / 4 skipped on a local merge with `main` (clean, 0 conflicts).


FlyM1ss:
## Review findings addressed (`038a49a`)

All five from the review, plus `main` merged in to clear the stale #104 failure the branch's base carried.

| # | Finding | Fix |
|---|---|---|
| 1 | Nothing alarmed on wholesale drift — the rename served the Phase-A feed for hours on per-item warnings alone | `get_latest_panel_payload` logs **ERROR** when a non-empty batch projects to zero entries. One bad story stays a warning; an empty batch is "no news", not drift |
| 2 | Items wire shape lived only as inline test dicts → drifted with the adapter, stayed green | `items-wire-fixture.json` records it once (key set verified against prod); happy path now driven from it; guarded against the retired `title`/`link` vocabulary |
| 3 | `schema_version` accepted but never read, doc called it the "evolution valve" | Doc now states plainly that no consumer gates on it and a v2 would degrade to fallback, not detection |
| 4 | New doc's status line said AF #359 was unmerged and named #107 as the consumer | Corrected: AF live, consumer is #110, with why #107 shipped without the adapter half |
| 5 | Spec still read `Status: design — awaiting review` | Marked implemented; rollout section kept as history with a postscript on why the order inverted |

### Verification

- **Full suite merged with `main`: 976 passed, 2 skipped, 0 failed.**
- **Live prod, healthy producer** → 50 real feed entries, `status=ok`, **0** drift errors.
- **Simulated the incident** (producer reverts to `title`/`link`) → panel stays alive on the Phase-A fallback and emits exactly **1** ERROR:
  > `items feed produced 0 usable entries from 5 raw item(s) — producer wire-shape drift? falling back to the Phase-A feed`
- Guards mutation-tested rather than trusted: an over-broad alarm fails `test_items_empty_list_does_not_log_drift_error`; renaming keys in the fixture fails **4** tests loudly.

### Known gap (documented, not fixed)

Nothing here can *catch* a producer rename — the producer is mocked, so AF and ATL can still only drift between deploys. The fixture makes the assumed shape reviewable and the ERROR makes a break visible within one poll. If this seam breaks a third time the answer is a canary against the real endpoint, not more mocked coverage.

FlyM1ss:
Cleanup pass + drift now visible in the UI. Two commits on top of the reviewed head.

**`c4bbfb1` — refactor.** Three findings, all in code this PR added:
- `_feed_from_items` hand-rolled the `{headline,source,url}` triple `_story_fields` already owns. The duplication is *new* — the rename is what made the two projections converge.
- The drift check was inline at one call site, but "non-empty raw projects to nothing" is a property of projection. `_representative_feed` had the identical shape and no alarm — and it's the fallback, so when *it* drifts the feed goes blank rather than merely stale. Now `_alarm_if_all_dropped`, applied to both.
- Rationale prose was stated twice in one function; now once, on the helper.

**`8185a6e` — badge.** The ERROR was a signal nobody subscribes to, which is the incident's own failure mode one layer up. Drift now sets `status: degraded` + a reader-facing `status_reason`; `home-news-signals.js` already renders it, so no frontend change.

| scenario | feed | badge |
|---|---|---|
| healthy | 6 | hidden |
| items drift | 2 (Phase-A fallback) | `degraded: news feed incomplete — upstream story format changed` |
| signals drift | 0 | same |
| producer degraded + drift | 2 | `degraded: 1 of 3 sources timed out; news feed incomplete — …` |
| quiet news day | 0 | hidden |

Row 2 is the 2026-07-14 incident — it now says so. The producer's own `status_reason` is appended to, never replaced.

**Worth flagging:** extracting the check moved its `raw and` guard out from under the `if items:` block that had made it unreachable, so that half needed coverage on the signals side — a mutation dropping `raw and` initially *survived*. Pinned by `test_empty_signals_artifact_does_not_log_drift_error`.

982 passed, 2 skipped. Mutations verified fatal: badge always-on, overwriting the producer's reason, dropping the signals-path drift, dropping the non-empty guard, and breaking `_story_fields` (15 tests — the sharing is enforced, not just documented).

PR #120, opened 2026-07-15
fix(news): guard the panel signals projection

### LINKED ISSUE #119: Panel signals passthrough has no drift guard
Found reviewing #117.

`get_latest_panel_payload` returns the wire`'s `signals` dict raw:

```python
signals = body.get("signals") or {}
...
"signals": signals,
```

Nothing projects it, so `_alarm_if_all_dropped` — which can only ever see what a projection dropped — is blind to it. It guards the two *feed* projections only.

Nothing server-side reads `sentiment_score`. The browser does:

```js
Number(s.sentiment_score).toFixed(2)   // -> "NaN" if the key is missing
```

So a producer rename of `sentiment_score` renders `NVDA NaN` in every chip with **no drop, no exception, no `degraded` badge, no ERROR log**. The feed keeps building (it reads `headline`/`url`/`source`/`published`), so the drift alarm cannot fire. That is the 2026-07-14 `title`/`link` outage one field over, and the `score` -> `sentiment_score` rename passed within inches of it.

Pre-existing; not introduced by #117. Fixed in #120.

FlyM1ss:
Closes #119.

The panel's `signals` dict reached the browser raw from the wire — the one panel path with no projection, and so the one path `_alarm_if_all_dropped` was blind to, since it only ever sees what a projection dropped. Nothing server-side read `sentiment_score`; the browser did, where `Number(undefined).toFixed(2)` renders the string `"NaN"`. A rename would have painted `NVDA NaN` into every chip with no drop, no exception, no badge, no log. The feed keeps building (different keys), so the existing alarm could not fire.

Project `signals` to the keys the panel renders, drop what it can't, run the existing alarm over it. Independent of the feed projection: a story with no sentiment read is still a story, so coupling them would let one break take out both.

**The alarm is appended *after* the feed block, not folded in above.** The items alarm assigns `drifted` outright rather than OR-ing it, so seeding it earlier is clobbered whenever the items feed loads — i.e. always, in prod. The guard would read correctly and fire never. `test_panel_signals_drift_survives_a_healthy_items_feed` pins that; the other drift tests leave `_http_get_items` unpatched and cannot see it.

**Behavior change:** an entry missing a rendering key is dropped rather than rendered broken, and an all-dropped signals list escalates to `degraded`. Verified against the producer that this cannot fire on a legitimate payload: `_PANEL_SIGNAL_KEYS` is a subset of the vendored v2 schema's per-entry `required` list, the producer builds all four unconditionally, and `_PUBLIC_STRIP` is top-level-only.

Second commit is self-review: mutation testing showed 3 of the 4 keys in the new guard were themselves unpinned (dropping `sentiment`/`url`/`source` left the suite green). Also the obvious non-dict test is vacuous — a `str` substring-matches, so it passes with the guard deleted; parametrized on `None`/`42` instead.

Stacked on #117 (same files). Suite 984 passed / 4 skipped.

https://claude.ai/code/session_01Pbs5wnQZhTi9CQ3pb14vDo

### DIFF
--- a/dashboard/backend/integrations/news_sentiment.py
+++ b/dashboard/backend/integrations/news_sentiment.py
@@ -342,7 +342,8 @@ def _feed_from_items(items) -> list:
     return feed
 
 
-def _alarm_if_all_dropped(raw, projected: list, *, source: str, consequence: str) -> bool:
+def _alarm_if_all_dropped(raw, projected: Union[list, dict], *, source: str,
+                          consequence: str) -> bool:
     """Whether `raw` projected to NOTHING — i.e. the wire contract moved rather
     than one story being off-spec — logging an ERROR if so. An empty batch is
     "no news", not drift.
--- a/dashboard/backend/integrations/news_sentiment.py
+++ b/dashboard/backend/integrations/news_sentiment.py
@@ -377,6 +377,44 @@ def _representative_feed(signals: dict) -> list:
     return feed
 
 
+# The keys home-news-signals.js reads off each `signals` entry to render a row.
+# `rationale` is deliberately absent — the panel null-coalesces it (`|| ''`) —
+# and so are headline/published/guid/n_articles, which only the feed projection
+# needs. Coupling the two projections' requirements would drop a usable story
+# from the feed because its sentiment read was broken, or vice versa.
+_PANEL_SIGNAL_KEYS = ("sentiment", "sentiment_score", "url", "source")
+
+
+def _project_panel_signals(signals: dict) -> dict:
+    """Panel `signals` projected to the entries the UI can actually render.
+
+    This dict used to reach the browser raw from the wire. That made it the one
+    panel path with no projection — and therefore the one path
+    `_alarm_if_all_dropped` was blind to, since it can only ever see what a
+    projection dropped. Nothing read `sentiment_score` before the browser did,
+    where `Number(undefined).toFixed(2)` renders the *string* "NaN": every chip
+    would read "NVDA NaN" with no drop, no exception, no badge and no log. That
+    is the 2026-07-14 title/link outage one field over, and the score ->
+    sentiment_score rename passed within inches of it.
+
+    Naming the missing key in the log is the point: "missing sentiment_score"
+    on every ticker at once reads as a rename, where a bare "malformed signal"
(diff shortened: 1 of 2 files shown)


FlyM1ss:
Note on CI: `ci.yml` triggers `pull_request` on `main` only, so the backend suite does **not** run on this PR while it is stacked on `chore/news-signals-v2-cutover` — the green checks above are Vercel only. It retargets to `main` automatically when #117 merges, and CI fires then.

Run locally against this branch in the meantime: `984 passed, 4 skipped` (full `dashboard/backend/tests/`).

FlyM1ss:
`/simplify` pass (4 review angles) — one substantive fix, pushed as f054180.

**Accumulation made uniform.** The items alarm assigned `drifted` while the other two OR-ed it, which is why the panel-signals alarm had to be appended last with a comment holding it in position. All three now OR, so each projection sits with its own alarm and order carries no meaning. Behaviour identical (`drifted` starts `False`).

**Worth knowing:** the obvious tidy-up — `drifted or _alarm(...)` — is a silent regression. `or` short-circuits, so once anything has drifted the later alarms are never *called* and never log. It reads fine and passed all 48 tests, so it's now pinned by `test_every_concurrent_drift_gets_its_own_alarm`. Alarm-first (`_alarm(...) or drifted`) is load-bearing, not stylistic.

Also: independence invariant stated once (at `_PANEL_SIGNAL_KEYS`), and dropped the incident retelling `_alarm_if_all_dropped` already carries.

Reviewed and rejected: generalizing the three projection+alarm pairs into one mechanism (three different trigger conditions and raw types — it'd move complexity, not remove it); deriving `_PANEL_SIGNAL_KEYS` from `signals-v2.schema.json` (test-only fixture; prod importing from `tests/` is worse).

**Follow-ups, not in this PR:**
- `_project_panel_signals` keeps the full raw entry (`projected[sym] = s`), so ~1–3KB of unread `headline`/`guid`/`n_articles` still ships to the browser. Safe to trim (no consumer reads them) but it changes the payload shape.
- `docs/integrations/finsearch-news-items.md` still describes the alarm as covering two projections; it's three now, and the traceability table has no row for the new guard.

Suite 998 passed / 2 skipped.

Issue #123, opened 2026-07-15
Wire RunManifest.news_sentiment_source (drift has no channel on the backtest path)


FlyM1ss:
Follow-up from #122 (deferred there deliberately, not overlooked).

`RunManifest.news_sentiment_source` (`api/v2/models.py:164`) is declared, flows through `runs.py:367`, and is **never assigned** — set to `None` in `BacktestBackend.__init__` (`backtest_backend.py:77`) and nothing writes it. It has been dead since `feat(v2): add BacktestBackend` (02102b8). Its intended job, per `agent-api-foundation-design.md:212`, is the per-run reproducibility manifest: record which news source fed this run.

**Why it matters now.** #122 adds a wholesale-drift alarm to the backtest signals projection, but its only channel is `logger.error`. The panel escalates drift to a `degraded` badge; the backtest has no equivalent. `_alarm_if_all_dropped`'s own docstring is blunt about why that is not enough:

> a log line only ever reaches whoever happens to be reading logs, and nobody was

So a backtest whose sentiment silently went empty mid-run is currently reconstructible only from server logs — a reproducibility gap in the field literally designed for reproducibility.

**Why it was not done in #122.** Escalating drift into `news_sentiment_source` while it is otherwise always `None` makes a provenance field cosplay as a status flag (`None` meaning "unknown", `"degraded"` meaning "drift seen"). The field's real job should land first; drift reporting rides on top.

**Open questions for whoever picks this up:**
- Provenance value shape — producer URL? artifact `as_of` + `schema_version`? Should be enough to re-fetch what the run actually saw.
- Run-level field vs per-step event: does drift on step 5 stick for the run, get clobbered, or need a count?
- Does it belong on `RunManifest` at all, or does drift want its own field so provenance stays provenance?

No wire-contract break either way: the field already exists as `Optional[str]` and consumers already handle `None`, so no `schema_version` bump.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #124, opened 2026-07-15
docs: correct the news drift-escalation claims

FlyM1ss:
Docs-only. The two FinSearch integration docs described a world with **one** drift check; after #120 and #122 there are four, across two entry points, and they disagree on what a drift is worth.

**Fixed**
- `status: degraded` was stated as a property of drift. It's a property of the *panel's* three call sites. `get_news_sentiment` (#122) discards the bool — a backtest step has no badge and no status field, so drift there is log-only pending #123. Now documented as a gap, not a design.
- "The drift check in `get_latest_panel_payload`" — singular, now false. `_story_fields` is shared, so a `headline` rename trips the panel *and* the backtest with different visibility.
- Both Traceability tables had no row for either new guard. Added, with test anchors.
- The `finsearch-news-sentiment.md` failure-mode table covered only transport failures (401/503/404/…) — nothing for the drop/alarm behavior that is now most of the fail-visible story.
- Both status headers were stale (#110 described as pending; the adapter described as unbuilt).

**Deleted: the reference sketch.** The adapter is built, so the sketch was a second copy of the truth — and it had drifted into demonstrating the exact defect #122 fixed: a projection loop with no per-ticker isolation, where one off-spec ticker raises out and the loader's fail-closed catch blanks *every* ticker. A reader following it reproduced #119/#122. Replaced with a where-it-lives table plus the two facts worth keeping (ISO-string coercion; ATL doesn't compute sentiment).

**Not stale, deliberately left:** "a rename of a field ATL *doesn't* read is invisible here". Its example is `editorial_score`, which `_feed_from_items` still never touches — the claim holds as written.

Verified against the merged code rather than review notes: all 4 call sites read, `_coerce_timestamp` confirmed to return `None` (not raise), `external_run_service.py:430-431` confirmed to still serialize ISO. No heading cited by code/tests was renamed. News suite 63/63.

### DIFF
--- a/docs/integrations/finsearch-news-items.md
+++ b/docs/integrations/finsearch-news-items.md
@@ -1,7 +1,7 @@
 # FinSearch Raw News Items — the `news-story v2` Contract (Phase B)
 
 **Audience:** both teams — AF producer (`Main/backend/api/signals_views.py`, Agentic-FinSearch) and ATL consumer (`dashboard/backend/integrations/news_sentiment.py`).
-**Status (2026-07-14):** **live.** AF PR #359 shipped the producer and is deployed (wire shape verified against prod). The ATL consumer lands in PR #110 — ATL PR #107 merged at a commit that predated the `headline`/`url` rename, so between the two merges the adapter read the retired keys and the panel silently served the Phase-A fallback; #110 is the fix.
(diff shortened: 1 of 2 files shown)


FlyM1ss:
Doc changes only, will directly merge. Major doc updates will happen after more work is done

PR #132, opened 2026-07-16
Add Daily Leaderboard to Competition

Allan-Feng:
## Summary
- Competition subtabs: **Daily Leaderboard** (first) · **Competition Leaderboard** · Participating Teams · About
- Daily board reuses the existing leaderboard chart/table via `GET /api/v1/leaderboard?period=daily` over the last completed US weekday; caches under `leaderboard-daily`
- Baselines auto-compute on load; LLM models refresh via `dashboard/scripts/refresh_daily_leaderboard.py --models` (nightly after market close)

## Test plan
- [ ] Competition → Daily Leaderboard shows same layout as Competition Leaderboard with Phase=Daily and a 1-day window
- [ ] Competition Leaderboard still shows the Apr–May preseason window
- [ ] `GET /api/v1/leaderboard?period=daily` returns `period: daily` and ranked entries
- [ ] `pytest dashboard/backend/tests/test_leaderboard_api.py -v` passes
- [ ] Optional: `python dashboard/scripts/refresh_daily_leaderboard.py` refreshes baselines

### DIFF
--- a/dashboard/backend/api/routers/leaderboard.py
+++ b/dashboard/backend/api/routers/leaderboard.py
@@ -15,15 +15,22 @@ router = APIRouter(prefix="/v1/leaderboard", tags=["leaderboard"])
 
 
 @router.get("")
-async def api_get_leaderboard(refresh: bool = Query(default=False)):
+async def api_get_leaderboard(
+    refresh: bool = Query(default=False),
+    period: str = Query(
+        default="contest",
+        description="Leaderboard period: 'contest' (fixed preseason window) or 'daily' (last completed weekday).",
+    ),
+):
     """
-    Official competition leaderboard for the configured contest window.
+    Official competition / daily leaderboard for the requested period.
 
     Baselines are computed from Alpaca hourly backtest data and cached in SQLite.
     Pass ?refresh=true to recompute (e.g. after config change).
+    Pass ?period=daily for the rolling one-day board.
     """
     try:
-        return get_leaderboard(force_refresh=refresh)
+        return get_leaderboard(force_refresh=refresh, period=period)
     except RuntimeError as exc:
         raise HTTPException(status_code=503, detail=str(exc)) from exc
     except Exception as exc:
--- a/dashboard/backend/domain/leaderboard/service.py
+++ b/dashboard/backend/domain/leaderboard/service.py
@@ -26,10 +26,11 @@ from dashboard.backend.domain.leaderboard.baselines import (
 )
 from dashboard.backend.domain.leaderboard.strategies._common import reference_start_date
 from dashboard.backend.domain.leaderboard.strategies import get_strategy
-from dashboard.backend.paths import CONFIG_DIR
+from dashboard.backend.paths import CONFIG_DIR, DATA_DIR
 
 LEADERBOARD_MODE = "leaderboard"
 VALID_PERIODS = ("contest", "daily")
+_SKIP_CACHE_PATH = DATA_DIR / "leaderboard_skip_cache.json"
 
 # H6 integrity threshold: an LLM entry must have decided at least this fraction
 # of its steps with the model itself. Below it, the curve is mostly a rule-based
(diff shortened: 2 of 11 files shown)


FlyM1ss:
Reviewed (backend correctness, H6 integrity, route-contract freeze, frontend/XSS). Two things to fix before merge; everything else is clean.

**Clean:** H6 guard is intact (the daily path still hard-skips `auto_compute:false` LLM strategies — traced end-to-end, not just from the diff). No route added/changed, so the route-freeze golden sets are untouched. No XSS introduced (new render code uses `textContent`; backend/frontend field names match).

### 1. Blocker — 4 failing tests (test-only, verified fix)

`test_service_move.py` was never updated for the new `ensure_leaderboard_runs` signature. Every real caller is correct — only the test's stub is stale. Two edits:

```python
# dashboard/backend/tests/domain/leaderboard/test_service_move.py:153
-        lambda force_refresh=False: {
+        lambda force_refresh=False, period="contest", config=None: {
```
```python
# same file, ~line 195 — get_leaderboard() now returns 4 more keys
     assert set(result.keys()) == {
         "window", "updated_at", "total_entries", "leader", "entries", "display_capital",
+        "board_title", "period", "standings_label", "phase_label",
     }
```

Verified locally on `0af3190` merged with current `main`: **1041 passed, 6 skipped, 0 failed**.

Note your branch predates #126/#133, so it needs a `main` update too — #133 (merged) fixes the 4 `test_protocol_api.py` failures you were also seeing.

### 2. Important — daily tab is the new default but boot CSS doesn't reveal it

The PR defaults the competition tab to `daily` (`app.html:27`, `:47`, plus the `app.js` defaults) and adds a `daily` arm to the subtab-*highlight* selectors — but the rule that actually reveals the panel still only matches `leaderboard`:

```
app.html:73  html[data-nav-boot][...][data-nav-competition-tab="leaderboard"] #leaderboardView { display: flex !important; }
```

So the **default** landing path is the one the anti-FOUC block no longer covers. Needs a `="daily"` arm (or a shared selector).

### Minors (your call)
- `js/leaderboard.js` ~785: contest subtitle hardcodes `'Sep 1 – Oct 30, 2026'` although `window.label` is in hand and is used correctly for the daily case.
- `js/leaderboard.js` ~801: `hiddenSeries = new Set()` moved from first-load-only into `loadLeaderboardData()`, so it now resets on *every* call — curve selections are discarded when re-navigating to an already-active board.

Not flagged: `daily_window_dates()` using UTC. It diverges from ET for ~4h/night, but I simulated it — the UTC-picked day is always already closed, and ET-based logic would pick the *staler* day. Current behaviour is fine.

FlyM1ss:
Pushed \`8999aec\` addressing the review findings:

- **FOUC (#1)** — boot CSS now reveals \`#leaderboardView\` for the \`daily\` default tab, so the board no longer flashes blank before \`app.js\` runs.
- **Skip-cache growth (#2)** — \`_prune_stale_window_skips\` drops earlier-window entries for the rolling daily session (no-op for the fixed contest window); unit-tested.
- **Race / non-atomic write (#3)** — the sidecar now writes to a unique temp file + \`os.replace\` (same dir → atomic), with a best-effort comment.
- **UTC boundary (#5)** — added the "intentionally conservative" note on \`daily_window_dates\`.
- **Style nit** — collapsed the stray blank line after \`resolve_leaderboard_config\`.
- **Copy (#4, partial)** — the daily About panel + config \`description\` no longer promise an auto-running nightly job; \`refresh_daily_leaderboard.py\`'s header now states it is **not scheduled** by anything in the repo.

Also fixed **pre-existing red CI on this branch**: \`test_service_move.py\` fixtures weren't updated for this PR's own \`config=\` kwarg / new \`get_leaderboard\` response keys (4 failures). Full backend suite now green — **1008 passed, 2 skipped**.

⚠️ **Follow-up not resolved in code (#4):** nothing schedules \`refresh_daily_leaderboard.py\`, so LLM entries won't appear on the daily board until a schedule is wired — and on free-tier Render the leaderboard DB is ephemeral, so an off-instance schedule won't reach prod without a persistent/shared DB. Worth confirming before merge.

FlyM1ss:
Filed the finding #4 follow-up (unscheduled daily refresh job) as #145 so it survives merge.

FlyM1ss:
Merged as 71f7b18. Remaining follow-up — the unscheduled daily refresh job — is tracked in #145. Prod deploy fires automatically once Backend tests pass on main.

PR #134, opened 2026-07-16
feat: persist agents & strategies in Postgres

FlyM1ss:
Optional Postgres backends for `external_agents`, `agent_versions`, and `strategies`, selected by `CONTENT_DATABASE_URL`, cloning the shipped `USERS_DATABASE_URL` users fix. Without them every registered agent and every issued API key dies on each Render deploy (free tier, no disk) — and `resolve_api_key()` is the sole auth path for `/api/v1` and `/api/v2`.

Each store module gained a `_build_*_store()` factory returning either a Postgres twin (`*_postgres.py` sibling) or the original SQLite store. Singleton names are unchanged, so there are no caller changes. Unset selects SQLite; set-but-unreachable fails loud rather than silently degrading. Each factory `print()`s its choice naming the host/db — `logger.info()` is invisible under the prod logging config, and a bare "postgres" cannot distinguish the intended database from a typo'd one. Credentials are never logged (`db_url.py`).

- Spec: `docs/superpowers/specs/2026-07-15-agent-strategy-persistence-design.md`
- Plan: `docs/superpowers/plans/2026-07-15-agent-strategy-persistence.md`

CI is green at **1087 passed / 6 skipped**, with the 13 `@pg_only` tests *running* against a real Postgres service rather than skipping — the skip count is the proof, since `skipif` fails open.

`CONTENT_DATABASE_URL` is already set on the prod service (same pooled Neon DB as `USERS_DATABASE_URL`), so the rollout order (env var first, then merge) is satisfied and merging activates the feature rather than shipping a no-op.

Follow-ups filed rather than folded in: #135 (twins have no lazy-migration path), #136 (`@pg_only` fixtures could wipe a real DB), #137 (five Postgres coverage gaps), #138 (CI is PG 16, prod is 18.4), #139 (pre-existing plot-test flake), #140 (run history still ephemeral), #141 (three small cleanups).


FlyM1ss:
Merged as `821cd56`. Recording what was verified and what was deliberately left, so the trail survives the PR going read-only.

**Pre-merge gate (checked at source, not from notes).** `CONTENT_DATABASE_URL` and `USERS_DATABASE_URL` are both set on the prod service, both well-formed, both pooled, and both point at the *same* Neon database — so the rollout order held and this merge activates the feature rather than shipping an inert no-op. A read-only `information_schema` query confirmed prod's schema contained only `users`/`auth_sessions`, so this deploy creates the three tables fresh and complete; there was no stale-shape hazard.

**Final review** ran four independent angles (cross-task seams, prod boot safety, security, test integrity). All four returned SHIP; 0 Critical. Everything they found that was worth fixing was fixed here rather than deferred:

- `4cd835f` — psycopg quotes the **entire** connection string, password included, into its error when the URL is malformed (anything not starting with `postgresql://` parses as a keyword DSN). The factories are deliberately try/except-free, so that traceback *is* the boot failure and the live Neon password would land in the deploy log. A scheme guard closes it: every malformed shape leaks, every well-formed one does not, so the check is the exact boundary and the fail-loud path is untouched. Applied to `users_postgres.py` too — the leak pre-dates this branch there and had the longest exposure.
- `1ce72b4` — `test_env_isolation` passed with the conftest pops deleted, and indeed with no conftest at all: it asserted the vars were *absent*, and they are never set on a runner. Now re-runs a named sibling in a subprocess under a hostile ambient value; mutation-tested (removing the pop turns the new test red while the two old ones stay green). Also pins `cash_allocation`, the only non-TEXT column, which no test ever wrote while `service.py` writes a float on every real registration.
- `00acbd3` — records that the twins have no lazy-migration path (#135).

**Verification worth trusting:** CI is green with the 13 `@pg_only` tests *executing* against a real Postgres, not skipping — `skipif` fails open, so the skip count (6, well under 13) is the proof, and `test_ci_provides_a_live_postgres` would go red rather than quiet if the wiring regressed.

**Follow-ups filed before merge**, most consequential first: #135 (twins have no lazy-migration path — three review angles converged on this), #136 (`@pg_only` fixtures would delete from whatever `TEST_POSTGRES_URL` names, with no allowlist), #137 (five Postgres coverage gaps), #138 (CI pins PG 16, prod Neon is 18.4), #139 (pre-existing `test_plot_png_cached_per_run` flake — unreproduced, not proven benign), #140 (run history is still ephemeral, so agents now outlive their own history), #141 (three small cleanups).

Next: confirm the deploy and read the four `<store> backend: postgres (<host>/<db>)` lines in the Render logs. `sqlite (ephemeral on Render)` there would mean the feature is inert.

PR #146, opened 2026-07-18
fix: harden the Postgres persistence tier

### LINKED ISSUE #135: Postgres twins have no lazy-migration path
Found by three independent angles of the #134 final review, converging.

The three Postgres twins added in #134 (`domain/agents/repository_postgres.py`, `domain/agents/version_repository_postgres.py`, `domain/strategies/repository_postgres.py`) contain **zero** `ALTER TABLE`. They fold the SQLite store's five accreted lazy migrations into their `CREATE TABLE`, which is correct for a table that starts empty — and leaves no mechanism for the *next* column.

**Failure:** add a column to both `_init_schema`s → `CREATE TABLE IF NOT EXISTS` no-ops on prod's existing table → column never appears → every query naming it raises `UndefinedColumn`, 500ing the whole agent surface. `/health` still passes, so there is no rollback signal.

**Why nothing catches it:** the SQLite tier is the default in tests, and CI's `postgres:16-alpine` service container is **empty on every run** — so the `@pg_only` tier only ever exercises the CREATE path, never the migrate path. Green CI, broken prod.

`users_postgres.py:52` already has the pattern (`ALTER TABLE users ADD COLUMN IF NOT EXISTS discord_user_id`, "Lazy migration for existing deployments"). Postgres supports `ADD COLUMN IF NOT EXISTS` natively.

#134 landed a comment in each `_init_schema` recording the constraint (00acbd3), which is a tripwire for a careful reader, not a fix.

**Options:** adopt the `ADD COLUMN IF NOT EXISTS` convention explicitly; or have the `@pg_only` tier run twice (once virgin, once against an already-created schema) so the migrate path is exercised; or a real migration tool.

Not urgent — nothing is broken today. It becomes urgent the moment someone adds a column.

### LINKED ISSUE #137: Postgres tier has five known coverage gaps
Gaps in the `@pg_only` tier from #134, collected from the per-task and final reviews. None is a live bug — the shipped SQL was verified correct by inspection in each case — but each is a mutation the tier would not catch.

1. **`claim_agent` / `reclaim_agent`** — the only 2 of `AgentStore`'s 14 methods with zero Postgres coverage.
2. **`scopes`** — no `@pg_only` assertion. `_public_agent`'s `or DEFAULT_SCOPES` would mask a wrong-but-non-empty stored default. This is an authorization surface.
3. **`last_used_at` touch-on-use** — `test_agent_key_lifecycle_postgres` asserts only `is not None`, and `create_agent` already sets it non-None, so a dropped touch-on-use `UPDATE` in `resolve_api_key` goes uncaught.
4. **Multi-row `ORDER BY`** — nothing exercises ordering on `list_agents` / `list_builtin_agents`, so an `ASC`/`DESC` swap goes uncaught.
5. **`nbytes` on the widen path** — the widen-after-20-collisions tests don't assert the argument passed to the mocked `token_hex`, so swapping the narrow (`_CODE_LENGTH // 2`) and widened (`_CODE_LENGTH`) call sites goes uncaught.

Context: #134 closed a related gap of exactly this shape — `cash_allocation`, the feature's only non-TEXT column, was never written non-NULL by any test while `service.py` writes a float on every real registration. Declaring it `TEXT` would have kept the entire tier green until the first prod registration. These five are the same class.

### LINKED ISSUE #138: CI tests Postgres 16, prod Neon runs 18.4
`.github/workflows/ci.yml` pins the `@pg_only` service container to `postgres:16-alpine`. The prod Neon database reports **PostgreSQL 18.4** (observed directly via a read-only `version()` query during the #134 pre-merge check).

The `@pg_only` tier exists specifically to catch Postgres dialect problems before they reach prod, so it testing a different major version than it ships to is a hole in exactly the thing it guards.

**No live risk today** — every construct the twins use is ancient and stable across both versions: `ON CONFLICT ... DO NOTHING` (9.5+), `CREATE INDEX IF NOT EXISTS` (9.5+), `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` (9.6+), `RETURNING *`, `DOUBLE PRECISION`, `TEXT`. Filed so the divergence is written down while it is cheap rather than after a version-specific difference bites.

Fix is a one-line image bump to match prod (`postgres:18-alpine`), ideally with a comment saying *why* the version is pinned to prod's.

FlyM1ss:
Batch B of the agent-persistence follow-ups — three issues, one tier.

- **Closes #135** — the agent twin gains a lazy-migration path (`ALTER … ADD COLUMN IF NOT EXISTS` for the 5 folded columns). Without it, a future column added only to `CREATE TABLE` no-ops on prod's existing table → `UndefinedColumn` 500s the whole agent surface while `/health` stays green. New `@pg_only` test recreates the pre-migration table to force the migrate path (RED before the fix, GREEN after). Scoped to the agent twin deliberately: it's the only twin with a folded migration history; version/strategy shipped complete and keep their documented convention.
- **Closes #137** — five `@pg_only` coverage gaps closed: claim/reclaim, raw `scopes` default (read straight from the column, since `_public_agent`'s `or DEFAULT_SCOPES` masks a wrong empty default), `last_used_at` touch-on-use, listing `ORDER BY`, and widen-path `nbytes`. Each was mutation-tested — inject the targeted bug, confirm the test goes red, revert.
- **Closes #138** — CI Postgres `16 → 18-alpine` to match prod Neon 18.x.

Verified on a local `postgres:18-alpine`: full suite **1113 passed / 5 skipped / 0 failed** with the live tier on.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,11 @@ jobs:
     # does not boot. Cheap insurance: the container starts in a couple of seconds.
     services:
       postgres:
-        image: postgres:16-alpine
+        # Pinned to prod's major version: the live Neon database runs
+        # PostgreSQL 18.x (checked via version() during the #134 pre-merge
+        # review). This tier exists to catch dialect problems before prod, so it
+        # must ship to the same major version prod runs -- see issue #138.
+        image: postgres:18-alpine
         env:
           POSTGRES_PASSWORD: test
           POSTGRES_DB: atl_test
--- a/dashboard/backend/domain/agents/repository_postgres.py
+++ b/dashboard/backend/domain/agents/repository_postgres.py
@@ -47,10 +47,14 @@ class PostgresAgentStore:
         # would raise UndefinedColumn -- 500ing this whole surface while /health
         # stays green. Nothing catches it first: the SQLite tier is the default
         # in tests, and CI's Postgres service container is empty on every run,
-        # so the @pg_only tier only ever exercises the CREATE path, never the
-        # migrate path. The columns below fold in the SQLite store's five lazy
-        # ALTERs because this table starts empty; that is why there is no
-        # migration here yet. users_postgres.py has the pattern to copy.
+        # so the @pg_only tier only ever exercises the CREATE path -- except
+        # test_agent_schema_lazily_migrates_an_old_table_postgres, which recreates
+        # the pre-migration table to force the migrate path. The five ALTER ...
+        # ADD COLUMN IF NOT EXISTS statements below re-add the columns the SQLite
+        # store accreted as lazy migrations, so a deployment whose table predates
+        # them is brought up to shape on the next boot; on a fresh or current
+        # table they no-op. Add the next column the same way (Postgres supports
+        # ADD COLUMN IF NOT EXISTS natively; users_postgres.py has the pattern).
         with self._get_connection() as conn:
             with conn.cursor() as cur:
                 # owner_user_id is deliberately a plain INTEGER with no FK to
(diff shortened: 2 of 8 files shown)


FlyM1ss:
Pushed `c999e96` — self-review follow-ups:

- **Migration test → populated-table case.** `test_agent_schema_lazily_migrates_an_old_table_postgres` now seeds a real legacy row before re-running `_init_schema`, then asserts the `ADD COLUMN … NOT NULL DEFAULT` backfilled `agent_type`/`scopes` on it (raw column read, bypassing `_public_agent`'s scopes masking). That's the actual prod scenario — a populated pre-migration table where a NULL `scopes` would be an authz hole — not just an empty table gaining columns.
- **Ordering test clock.** Swapped the fixed 6-item iterator for an unbounded `itertools.count`, so a change in `create_agent`'s `_utcnow_iso` call count surfaces as a clean assertion failure instead of `StopIteration`.
- **Doc version consistency.** Bumped the remaining `postgres:16-alpine → 18-alpine` refs in this workstream's plan/spec docs to match the shipped CI (#138).

Verification: full non-pg suite green locally (**1095 passed / 23 skipped / 0 failed**). The `@pg_only` assertions weren't run locally (no Docker engine reachable here) — **CI's pg18 tier is the live gate** for them.

Out of scope, flagging: `docs/superpowers/plans/2026-07-08-user-account-persistence-fix.md` still has two `postgres:16-alpine` refs, but that's the earlier user-account workstream's doc — left untouched.

PR #147, opened 2026-07-18
chore: three cleanups around Postgres persistence

### LINKED ISSUE #141: Three small cleanups around the Postgres persistence work
Three unrelated small things noticed during the #134 review. Grouped because each is a few minutes.

**1. The seed `backtest.db` carries leftover test-repro agents.**
`dashboard/storage/data/backtest.db` has exactly 2 rows in `external_agents`, both junk from an ownership-security repro:

| agent_id | name | owner_user_id | owner_browser_session |
|---|---|---|---|
| `agent_43b44a048b62` | `owner-a-repro` | NULL | NULL |
| `agent_59eac17648ac` | `intruder-b-repro` | NULL | NULL |

Both are ownerless and their API keys are unrecoverable, so they are inert — but they are committed test debris, the same class #106 cleaned out of this file. (`agent_versions` and `strategies` are both empty.) After #134 they stop being served in prod anyway, since the Postgres tables start empty.

**2. `render.yaml` documents `CONTENT_DATABASE_URL` but not `USERS_DATABASE_URL`.**
#134 added a `CONTENT_DATABASE_URL` entry (`sync: false`) to the `envVars` list. `USERS_DATABASE_URL` has never been listed, despite being set on the service and equally load-bearing — an operator reading `render.yaml` would not learn it exists. Current keys: `PORT`, `PYTHONUNBUFFERED`, `PYTHON_VERSION`, `DATABASE_PATH`, `CONTENT_DATABASE_URL`, `DISCORD_*`, `PUBLIC_APP_URL`.

**3. The persistence plan's own verification step cries wolf.**
`docs/superpowers/plans/2026-07-15-agent-strategy-persistence.md` (Task 9, Step 5) says to run:

```bash
timeout 10 python -m uvicorn dashboard.backend.app:app --port 8123 2>&1 | grep "backend:"
```

and to treat zero matches as "someone converted `print()` back to `logger.info()` — a blocking regression".

That command prints **zero lines on perfectly healthy code**. A pipe is block-buffered (a TTY is line-buffered), so the four short lines sit in the buffer and uvicorn's SIGTERM shutdown discards them. Confusingly, "Application startup complete" *does* appear, because that's uvicorn's logger writing to stderr — so it looks exactly like the regression it warns about.

Prod is unaffected for two independent reasons: `render.yaml` sets `PYTHONUNBUFFERED: true`, and Render captures logs from a pty. Verified: with `PYTHONUNBUFFERED=1`, or to a pty via `script(1)`, all four lines appear; redirected/piped without it, zero.

A verification step that fails open on healthy code is worse than none — it teaches the next reader to ignore it.

FlyM1ss:
Batch C of the agent-persistence follow-ups — three unrelated few-minute cleanups (grouped as filed in #141).

- **Seed DB** — dropped 2 leftover repro agents (`owner-a-repro`, `intruder-b-repro`) from `backtest.db` `external_agents`. Not purely cosmetic: both rows were **unowned** (`owner_user_id` and `owner_browser_session` NULL) and carried **full scopes** (`agents:register,runs:write,context:read,decisions:write,runs:read`) — and on the free tier this committed DB *is* the running prod DB. Only a sha256 hash and an 11-char prefix were ever exposed, so real risk was low, but these are credentials rather than debris.
- **render.yaml** — documented `USERS_DATABASE_URL` (`sync: false`) next to `CONTENT_DATABASE_URL`. It's set on the service and equally load-bearing, but an operator reading the yaml wouldn't have known it exists. Documentation only — prod doesn't sync from this file (`sync: false`, so no secret committed).
- **Plan doc** — fixed the Task 9 Step 5 verification that cried wolf. Piping uvicorn stdout to `grep` block-buffers the factory `print()`s and `timeout`'s SIGTERM discards them, so it showed **zero** `backend:` lines on healthy code — while "Application startup complete" still appeared, making the false alarm convincing. Added `PYTHONUNBUFFERED=1` (what `render.yaml` sets in prod). Verified at source: **4 lines with it, 0 without.**

Closes #141. Rebased onto `main` post-#146; suite green on that base: **1107 passed / 23 skipped**.

### Review fix (v2)

The first push of the seed-DB commit silently carried an **unrelated schema mutation**: `users` gained a `discord_user_id TEXT` column plus `idx_users_discord_user_id`, and the single `users` row was rewritten with a trailing `NULL`. Cause is the gotcha CLAUDE.md warns about, via a path worth naming — `users.py` migrates that column with a lazy **`ALTER TABLE ADD COLUMN`**, so a prior local app run left it pending in the untracked `backtest.db-wal` sidecar, invisible to `git status`; the `PRAGMA wal_checkpoint` used to prepare the commit then folded it into the main file, where it was indistinguishable from the intended row deletion inside a binary blob.

Functionally harmless (`users.py` guards with `if "discord_user_id" not in columns`, and prod's accounts are on Postgres via `USERS_DATABASE_URL` anyway) — but it made the diff wider than the PR claimed, in a file no reviewer can read.

Rebuilt the blob from the pristine `main` version instead of from the working tree. Verified: `sqlite3 .dump` of the committed blob differs from `main` by **exactly the 2 `external_agents` INSERTs and nothing else** — schema, indexes, and all 13 other tables identical; `PRAGMA integrity_check: ok`.


FlyM1ss:
Review pass done; one real finding, fixed and force-pushed (`dabd0f0` → `ad074ef`, also rebased onto `main` post-#146).

**Fixed:** the seed-DB blob carried an unrelated `users.discord_user_id` column + index that no reviewer could have seen in a binary diff — a lazy `ALTER TABLE` from a prior local run, pending in the untracked `-wal` sidecar and folded in by the checkpoint. Rebuilt from the pristine `main` blob; the committed blob's `sqlite3 .dump` now differs from `main` by exactly the 2 `external_agents` INSERTs and nothing else, `integrity_check: ok`. Details in the PR body.

**Verified, not changed:**
- The `PYTHONUNBUFFERED=1` plan fix reproduces exactly as described — 4 `backend:` lines with it, 0 without, "Application startup complete" present in both. The false-alarm mechanism is real.
- `render.yaml` parses clean, no duplicate keys, and it does already set `PYTHONUNBUFFERED` — which is what makes the plan recipe match prod.
- No reference to `owner-a-repro`/`intruder-b-repro` anywhere in the tree; `defaults.json` points at `agent_runs` (17 rows, untouched), not `external_agents`. Removal is safe.

**Rebase question, settled:** #146's edits to the shared plan doc are all `postgres:16→18-alpine` swaps; this PR's edit is Task 9 Step 5, ~24 lines away. No textual *or* semantic conflict — the rebase above was clean and needed nothing.

**Follow-up filed as #155** (out of scope here): the seed DB still ships a real user account + session row. The session is expired, so no live exposure — but it's the same class of junk this PR removes from `external_agents`. That issue also proposes documenting the `ALTER`-via-WAL trap in CLAUDE.md, since the existing warning only covers `CREATE TABLE IF NOT EXISTS`.

PR #163, opened 2026-07-22
Decouple Discord backtest delivery from long-running execution

Allan-Feng:
## Summary
- Discord `/backtest` now ACKs immediately and persists a notify job instead of blocking the interaction while polling (which dropped charts after ~11 minutes / Discord’s ~15m token).
- A background watcher posts metrics + equity PNG **in-channel** (with an @mention) when the API job finishes; open jobs resume after bot restart.
- `/backtest/run` returns a stable `live_run_id` so the notifier can key the exact run.

## Notes
- Final results are **channel-visible** (not ephemeral) by design so delivery survives long LLM runs.
- Job rows live in a bot-local SQLite file (`DISCORD_JOBS_DB` / `dashboard/storage/data/discord_backtest_jobs.db`). On Render workers the disk is ephemeral: a redeploy mid-run can lose the notify job (the backtest itself still completes on the API). Acceptable for now; durable Postgres notify queue can come later if needed.
- Bot needs **Send Messages** + **Attach Files** in the target channel.

## Test plan
- [ ] `pytest dashboard/backend/tests/integrations/test_discord_jobs.py dashboard/backend/tests/integrations/test_discord_wiring.py dashboard/backend/tests/domain/chat/test_discord_bot.py -q`
- [ ] Redeploy/restart Discord bot worker with this branch
- [ ] `/backtest` (or strategy **Run backtest**) → immediate “queued” ACK
- [ ] When run finishes → channel message with @mention, metrics, chart, Dashboard link
- [ ] Confirm bot role has Send Messages + Attach Files in that channel

### DIFF
--- a/dashboard/backend/api/routers/backtests.py
+++ b/dashboard/backend/api/routers/backtests.py
@@ -612,7 +612,23 @@ async def run_backtest_endpoint(
     # Mint run id before the worker starts so callers (Discord job watcher)
     # can key notifications on a stable id from the HTTP response.
     live_run_id = f"agent_{datetime.now().strftime('%Y%m%d_%H%M%S')}_{uuid.uuid4().hex[:8]}"
-    
+
+    # Publish "running" synchronously — before thread.start() — so a status poll
+    # landing in the gap between this return and the worker's first line cannot
+    # read a PRIOR run's completed state (running=False, runs_count>0) and report
+    # it as this run's result. Clearing runs_count/error retires the previous
+    # run's terminal signal; setting live_run_id lets the watcher key on this
+    # exact run. (PR #163 completion-detection race.) Item assignment on this
+    # global dict is atomic under the GIL, same as the worker's own writes.
+    global backtest_session_id
+    backtest_status["running"] = True
+    backtest_status["error"] = None
+    backtest_status["runs_count"] = 0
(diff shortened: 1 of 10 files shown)


FlyM1ss:
Pushed `beebd94` addressing the review findings on the async delivery path:

- **Completion-detection race (Medium):** the endpoint now publishes `backtest_status["running"]=True` *synchronously* before `thread.start()` (also clearing `runs_count`/`error`, setting `live_run_id`). This closes the window where a status poll in the gap before the worker's first line could read a *prior* run's completed state and post it as this run's result.
- **Watcher fallback hardening:** `/runs/latest/metrics` is now identity-gated — accepted only when it *is* this run (ids match) or when there's no `live_run_id`. Since Discord sessions are stable per user, "latest" could otherwise be a prior same-session run; the watcher now soft-fails rather than posting stale numbers.
- **At-least-once delivery:** documented at the notify site (post → mark NOTIFIED; a crash in the gap re-posts on resume — duplicate, never lost).
- **Dead `meta_json` column** dropped from the new (never-deployed) table.
- **Tests:** added `test_discord_watcher.py` — behavioral coverage for `watch_and_deliver_backtest` (happy path, the identity-gated fallback regression, accept-when-matching, error status).

Suite green locally: 1116 passed, 23 skipped.

PR #164, opened 2026-07-22
Fix My Agents links to open the latest backtest

Allan-Feng:
## Summary
- Clicking **1 backtest** / **View All Runs** on My Agents now jumps straight to Playground → Backtest with the agent's latest run selected.
- Root causes: navigation awaited a potentially slow `activate` ping, and `applyActiveAgent` cleared the pinned `selected-backtest-run-id` on same-session re-activate.
- Home agent card also renders/wires the same CTAs.

## Test plan
- [ ] Hard-refresh `/app` (`app.js?v=20`)
- [ ] My Agents → click **1 backtest** → lands on Backtest tab with latest run chart/metrics
- [ ] Click **View All Runs** → same behavior
- [ ] Home → My Agents module CTAs also open latest backtest
- [ ] Switching to a different agent still clears the previous selected run

### DIFF
--- a/dashboard/frontend/app.html
+++ b/dashboard/frontend/app.html
@@ -1496,7 +1496,7 @@
     <script src="js/agent-editor.js?v=7"></script>
     <script src="app.js?v=20"></script>
     <script src="js/leaderboard.js?v=13"></script>
-    <script src="home-page.js?v=32"></script>
+    <script src="home-page.js?v=33"></script>
     <!-- Load order: must come after app.js (API/API_BASE/escapeHtml) and
          market-events/marketEventRelativeTime.js (formatMarketEventRelativeTime),
          both loaded above — not beside the market-events/* block, which loads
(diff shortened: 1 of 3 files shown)


FlyM1ss:
Pushed `7168a08` addressing the review findings:

- **Cache-bust collision (Medium):** both this PR and #165 bumped `app.js` to `?v=21` with different content. Git would auto-merge that line silently, then serve stale JS to browsers that cached one version's `app.js?v=21`. Bumped this PR to **`?v=22`** so the two are distinct — and so the shared line surfaces as a *visible merge conflict* rather than a silent auto-merge. (`home-page.js` is fine: v33 here vs v32 on #165.) Whichever of #164/#165 lands second should keep its number distinct.
- **Redundant double `applyActiveAgent`:** added comments at `openAgentInBacktest`/`openAgentInPaper` explaining the first call is for immediate navigation and `activateAgent` re-applies (idempotent) + fires the deliberately-unawaited server ping.

`node --check app.js` passes.

Issue #175, opened 2026-07-23
feat: allocate / reclaim portfolio cash to agents (delete returns funds)


Allan-Feng:
### Goal

Let users **move money** between the account portfolio and agents, and get it back when an agent is deleted.

Depends on #174.

### Must achieve

- **Allocate**: unallocated → agent (cannot exceed available cash)
- **Reclaim**: agent → unallocated
- **Delete agent**: reclaimable cash returns to the account
- Portfolio / agent UI reflects the real allocations
- Agent money is traded only via **paper trading**; **backtest does not** move the account ledger

### Context

Split from #173

Allan-Feng:
Depends on create portfolio: #174. Epic: #173.

FlyM1ss:
Two representation/UX decisions surfaced while reviewing #174's implementation (PR #177). Both are deliberately **not** settled there, because they only become real once allocate/reclaim exists — recording them here so they aren't lost with that PR.

**1. Money is floating point on both halves of the ledger.**
- `user_portfolios.equity` / `.cash_available` — `REAL` (SQLite) / `DOUBLE PRECISION` (Postgres)
- `external_agents.cash_allocation` — `REAL`

Allocate/reclaim is where float drift starts to matter (repeated add/subtract round-trips, and "reclaim everything" needing to land exactly on zero). If this moves to integer cents, **both** halves have to move together — converting one and not the other is worse than leaving both as-is.

**2. The signed-in allocation donut is hardcoded to 100% Unassigned.**
`dashboard/frontend/js/portfolio.js:367` — *"Until allocate (#175) wires the ledger, signed-in pie is 100% Unassigned."* Correct as a placeholder, but note it currently **hides** `cash_allocation` values that users can already set through the agent editor (#165). Wiring the real split is part of this issue's "Portfolio / agent UI reflects the real allocations".

PR #165, opened 2026-07-22
feat: My Agents configure-and-run rebuild

FlyM1ss:
Demo 1 ("configure & run"): a non-technical user opens My Agents, sees a default agent, presses Configure, sets money + one plain-language instruction + a model, and runs a backtest — unaided.

**Frontend**
- My Agents grouped into **Foundation Models** / **External Agents** rows (type-filter dropdown removed); "Connect your own agent" placeholder reserves the External row.
- First visit auto-provisions a default Foundation agent (DeepSeek V4 Pro, seeded starter instruction, "Default" badge; localStorage-guarded so deletion doesn't resurrect it).
- First-class **Configure** button on every card (replaces ⋮ → Edit); ⋮ gains "Set as default".
- Editor now has **Simple** (capital + one instruction + editable model) ⇄ **Advanced** (legacy multi-step pipeline, unchanged). The instruction is stored as a 1-step `simple_instruction` pipeline reusing the proven `orders` output contract — no new run-time wiring.

**Backend** (no new routes — freeze guards green; no migration)
- `PATCH /api/v1/agents/{id}` accepts optional `model_name` (SQLite + Postgres twins).
- `MAX_AGENT_CASH_ALLOCATION` 3,000 → 1,000,000 (default stays 1,000); frontend mirrors updated.

**Verification**: backend suite 1109 passed / 24 skipped; Playwright end-to-end click-through of the exit-criterion path (provisioning, Simple save/reload persistence incl. model round-trip, mode toggles, replace-warning, backtest run, `?demo=1` regression).

Notes: auto-provisioning keys on browser identity (demo visitors aren't logged in) with a client-side guard — server-side cap is a candidate follow-up. The new PG `update_agent` parity test is `@pg_only` (skips without `TEST_POSTGRES_URL`); the SQL delta mirrors the SQLite twin verbatim.

Spec: `docs/superpowers/specs/2026-07-22-my-agents-configure-run-design.md` · Plan: `docs/superpowers/plans/2026-07-22-my-agents-configure-run.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Ya39uNCr9z5JMJ94eNbYgo

PR #177, opened 2026-07-23
feat: account-bound $10k portfolio (replace SAMPLE DATA)

### LINKED ISSUE #174: feat: create account-bound $10k portfolio (replace SAMPLE DATA)
### Goal

Give every signed-in user a **paper money** **\$10,000 portfolio** and stop showing SAMPLE DATA.

### Must achieve

- Signed-in users start with **\$10,000** unallocated cash
- Signed-in My Portfolio loads from **`GET /api/v1/portfolio`** . Do not render `PORTFOLIO_MOCK` / SAMPLE DATA for signed-in users
- Show **unallocated / cash available** correctly
- Guest / demo can keep the mock

### Not in this issue

Allocate / reclaim / delete-returns → #175

### Context

Split from #173

Allan-Feng:
## Summary
- Add account-bound portfolio ledger (`user_portfolios`) with `GET /api/v1/portfolio` for signed-in users; bootstraps at **\$10,000** unallocated cash.
- Persist via content DB (SQLite locally; Postgres when `CONTENT_DATABASE_URL` is set — same pattern as agents).
- My Portfolio UI reads the API for signed-in users and hides SAMPLE DATA; guests keep the mock.

Closes #174.

## Test plan
- [ ] `pytest dashboard/backend/tests/test_portfolio_api.py dashboard/backend/tests/test_app_composition.py -q`
- [ ] Sign in locally → open My Agents → My Portfolio shows \$10,000 Cash Available, no SAMPLE DATA badge
- [ ] `GET /api/v1/portfolio` with Bearer token returns equity/cash_available; second GET is idempotent
- [ ] Guest / logged-out still sees SAMPLE DATA mock
- [ ] With `CONTENT_DATABASE_URL` set, startup log shows `portfolio_store backend: postgres (...)`


FlyM1ss:
Reviewed and pushed fixes in `ab619c1` (plus a `main` merge in `e842e7b` — the branch was behind and conflicting). CI green; `test_portfolio_store_postgres.py .........` ran 9/9 against CI's pg18, no skips.

**Fixed**
- **No Postgres-tier test.** Prod sets `CONTENT_DATABASE_URL`, so the twin was the only implementation users hit and no line of it ran in CI. Added `test_portfolio_store_postgres.py` (bootstrap, per-user isolation, the `UniqueViolation` race, SQLite/PG payload parity); destructive fixture calls `require_local_postgres_url`.
- **`temp_stores` never patched the auth module.** `api/auth.py` binds `user_store` at import, so signup and `get_current_user` used the process-global store — the temp `users.db` sat unused and accounts leaked into the session DB, making the fixed email an ordering hazard. Rebound + guard test added.
- **`_init_schema()` per read/write** — 2 connections per GET, 6 on bootstrap, half of them DDL. Now `__init__` only, matching every other store.
- **Missing ALTER warning** in both tiers: `CREATE TABLE IF NOT EXISTS` no-ops on an existing table, so a column added for #175 would never reach prod (#135).
- **`app.js?v=26`** — `main` shipped v25 and this branch edits `app.js`; v25 would serve a stale cached copy.
- Empty holdings donuts drew no arcs (read as a failed chart); `assert` → `raise` (`-O` strips asserts); per-user isolation covered at store + API; CLAUDE.md lists portfolios as the fourth content store.

New tests are mutation-verified (removing the auth patch, ignoring `current_user`, and dropping the `owner_user_id` predicate each fail them).

**Not changed, deliberately**
- **Money stays `REAL`/`DOUBLE PRECISION`.** Switching to integer cents here while `external_agents.cash_allocation` stays float would put #175's two halves on different representations. Worth deciding for both together — flagging for whoever picks up #175, since float drift will surface once allocate/reclaim cycles land.
- **Agent donut is 100% Unassigned for signed-in users** (existing `// Until allocate (#175)` comment). Correct given the ledger has no allocation column, but it does hide `cash_allocation` values set via #165 — confirm that's the intended demo behavior.
- `GET` bootstraps a row. Idempotent, so safe; noting only that POST-to-bootstrap is the more conventional split.

No prod migration needed: `CONTENT_DATABASE_URL` is already set, so `user_portfolios` auto-creates in Neon at first boot and the startup log gains a fifth `... backend: postgres` line.

Docs gaps filed as #188 rather than grown into this PR.

FlyM1ss:
Merged as `3af0457`.

**GitHub Advanced Security — all 8 CodeQL alerts on this PR cleared before merge:**

| Alerts | Rule | Action |
|---|---|---|
| 1082, 1093, 1094–1097 | `py/import-and-import-from` | **Fixed** in `e84ecf0` — the two test files imported a module both as `import x.y as m` (needed, since `monkeypatch` patches attributes on the module object) and as `from x.y import Cls`. Standardised on the module-alias form; no behaviour change. |
| 1076, 1092 | `py/cyclic-import` | **Dismissed (won't fix)** — deliberate: `repository.py` imports the Postgres twin *inside* `_build_portfolio_store()` so `psycopg` only loads when `CONTENT_DATABASE_URL` is set, and the twin imports shared helpers back. Same cycle as the agents / agent-version / strategies / users stores. |

All were `note` severity with no security severity, and the CodeQL check was already passing — none gated the merge.

Verified on the merge ref after the fix: **zero open code-scanning alerts**, and the Postgres tier genuinely executed rather than silently skipping (`test_portfolio_store_postgres.py .........`, 9/9 — its `skipif` fails open, so green CI alone would not have proved it). Suite: **1175 passed, 7 skipped**.

**Follow-ups** (a merged PR is write-only, so these live elsewhere):
- #188 — docs for the account-bound $10k portfolio
- #175 — [comment](https://github.com/Open-Finance-Lab/AgenticTrading/issues/175#issuecomment-5055983176) recording the two decisions deferred out of this PR: float-vs-integer-cents across **both** halves of the ledger, and the donut currently hardcoded to 100% Unassigned (which hides `cash_allocation` values already settable via #165)

Unrelated to this PR: 3 open HIGH Dependabot alerts on `mcp` are **stale** — every advisory's vulnerable range excludes `1.28.1`, which `main` already pins via #176. Left open for Dependabot to re-scan rather than dismissed, since no dismissal reason would be truthful for an alert that is actually fixed.

PR #179, opened 2026-07-23
fix: make browser Back undo in-app dashboard navigation

### LINKED ISSUE #178: fix: browser Back does not undo in-app navigation
## Problem

In the dashboard SPA (`/app`), clicking primary nav / playground & competition subtabs changes the visible page, but the browser **Back** button does not return to the previous view. Forward is equally useless. Users expect Back to undo the last in-app navigation step (Home → Playground → Competition, etc.).

## Root cause

Navigation is entirely client-side DOM toggling via `navigateToPage()` / `switchPlaygroundTab()` / `switchCompetitionTab()` in `dashboard/frontend/app.js`:

1. **No `history.pushState`** on user navigation — only a few auth/deep-link cleanups call `history.replaceState`, which never creates a stack entry.
2. **No `popstate` listener** — even if the URL changed, Back would not restore the matching view.
3. Tab state is persisted in `localStorage` (`nav-state`) for refresh restore, which is orthogonal to the session history stack.

`resolveInitialNavigation()` already understands `?view=` / `?mode=` / hash deep links (`home`, `backtest`, `paper`, `contest`, `my-algo`, …), but those params are not kept in sync when the user clicks nav.

## Expected behavior

- Each distinct page/subtab change pushes a history entry (or replaces when landing / same destination).
- Browser Back / Forward restores the previous / next view.
- Refresh still lands on the current view (URL + existing `nav-state` fallback).
- Existing deep links (`?view=backtest`, Discord `?agent_id=&run_id=`) keep working.

## Suggested fix

In `dashboard/frontend/app.js`:

1. Map `{ page, playgroundTab, competitionTab }` ↔ `?view=` (extend the map for competition subtabs that currently have no view token).
2. On user-initiated nav: `history.pushState(state, '', urlWithView)`.
3. On boot / deep-link cleanup: `replaceState` so the current entry carries state without adding a duplicate.
4. On `popstate`: call `navigateToPage(...)` with a "don't touch history" option.
5. Skip push when the destination equals the current nav state.

## Acceptance

- [ ] Home → Playground → Competition, then Back twice returns through Playground to Home.
- [ ] Playground Agents → Backtest → Paper, Back restores prior subtabs.
- [ ] Competition Daily → Leaderboard → About, Back works.
- [ ] Hard refresh on a navigated view keeps that view.
- [ ] `?view=backtest` and Discord agent/run deep links still open the backtest playground.

Allan-Feng:
## Summary
- Wire `history.pushState` / `replaceState` + `popstate` into dashboard nav so browser Back/Forward restore the previous page/subtab.
- Keep `?view=` in sync with the visible page (including competition subtabs); boot/deep-link paths use `replace` so they don't pollute the stack.
- Fixes #178.

## Test plan
- [ ] Open `/app`, go Home → Playground → Competition; Back twice returns through Playground to Home.
- [ ] On Playground, switch Agents → Backtest → Paper; Back restores prior subtabs.
- [ ] On Competition, switch Daily → Leaderboard → About; Back works.
- [ ] Hard refresh on a navigated view keeps that view (`?view=` present).
- [ ] `?view=backtest` and Discord `?agent_id=&run_id=` deep links still open the backtest playground.
- [ ] Forward after Back restores the view you left.


FlyM1ss:
Pushed `ababd2c` (on top of the `main` merge in `2fe8748`) addressing the review findings:

| # | Finding | Fix |
|---|---|---|
| 1 | `?view=` map duplicated across `app.html` + `app.js` | One object: the boot script owns it and exposes `window.NAV_VIEW_MAP`. Also collapsed a **third** copy in `switchMode()`. |
| 2 | `popstate` registered at the end of a long async boot | Moved to `applyInitialNavigation()`; also guarded the `applyAgentRunDeepLink()` await so a failed deep link no longer kills every nav button. |
| 3 | Active-subtab guards removed the re-click refresh | Dropped both. `syncNavigationHistory` already refuses a duplicate push, so the stack stays clean. |
| 4 | Signed-out deep link left `agent_id`/`run_id` in the URL | Parked in `sessionStorage` and stripped from the URL. They were riding along on every later `pushState`, and `resolveInitialNavigation` prefers them over `?view=` — so a refresh from any page snapped back to the backtest tab. |
| 5 | Daily board emitted `?view=contest` | Emits `?view=daily`; `contest` still resolves as an alias. |
| 6 | `NAV_VIEW_MAP` / `viewParamForNavState` are hand-maintained inverses | Invariant documented on the function. |

**Verified**

- 15/15 browser checks (your test plan + one per finding), 0 uncaught JS exceptions.
- Map round-trips both directions across all 14 slugs; 10 distinct states, all linkable.
- Re-clicking the active subtab fires 2 API calls (was 0 with the guard) — the refresh affordance is really back.
- Backend suite 1167 passed / 30 skipped.

`app.js` cache-buster bumped to `v=28`.

The one CodeQL alert on this branch (`js/xss-through-exception`, `app.js` algo-chat) is **pre-existing** — byte-identical in `main`, which already tracks the same sink as `js/xss-through-dom`. It surfaced here only because the insertions shifted the line numbers. Not a blocker for this PR.

PR #180, opened 2026-07-23
feat: allocate / reclaim portfolio cash (delete returns funds)

### LINKED ISSUE #175: feat: allocate / reclaim portfolio cash to agents (delete returns funds)

Allan-Feng:
## Summary
- **Allocate / reclaim** between account unallocated cash and `agent.cash_allocation` (create, PATCH, and `POST /api/v1/portfolio/allocate|reclaim`).
- **Delete agent** returns reclaimable sleeve cash to the account ledger.
- Portfolio UI refreshes from the ledger and shows per-agent slices again.
- Backtests remain simulation-only (do not move the ledger). Paper per-agent sleeve isolation is a follow-up (shared Alpaca today).

Depends on #177 (#174). Closes #175.

## Test plan
- [ ] `pytest dashboard/backend/tests/test_portfolio_allocate.py dashboard/backend/tests/test_portfolio_api.py -q`
- [ ] Sign in → create agent with \$1k → Cash Available drops by \$1k; pie shows agent slice
- [ ] PATCH cash up/down moves unallocated correctly; over-allocate returns 400
- [ ] Delete agent → cash returns to unallocated
- [ ] Guest create (no login) still works without ledger debit


FlyM1ss:
Pushed review fixes as `7e99f7c` (plus a `main` merge — the `app.html` cache-bust conflict is resolved, `app.js?v=30`).

**Blocking bug this fixes.** Every agent that exists today holds a `cash_allocation` that never debited a ledger (#177 shipped read-only, and `reclaim_on_session_match` keeps minting more by claiming guest agents). Their owner's portfolio bootstrapped at the full $10k, so the same money was counted twice:

- `DELETE` credited the sleeve back, overshot equity, and raised **out of the route after the agent row was already deleted** → HTTP 500, agent gone.
- `PATCH`ing such an agent's allocation *down* → 400, so it could never be lowered.
- The sleeve could be spent a second time via create/allocate.

Reproduced against the branch before fixing; all three now have regression tests.

**The change.** Agent sleeves become the source of truth; `cash_available` is derived (`service._reconcile`) on read and after each write. Those states self-correct on the next read, `DELETE` becomes a write that cannot fail, and the compensating-transaction branches disappear — one write per transfer instead of two-plus-rollback.

**Also fixed**
- `PATCH` validates before any write and moves the ledger after the other fields land — a failed update no longer leaves money moved behind an error response.
- Per-user serialisation of transfers (check-then-write spans two tables, so the DB can't do it). Process-local; `_reconcile` keeps a multi-worker over-allocation *visible* rather than silent.
- `BEGIN IMMEDIATE` + busy timeout on the SQLite writer — concurrent writes were raising `database is locked`.
- `assert` → `RuntimeError` in both store twins, matching the siblings #177 already converted (asserts vanish under `-O`).
- Live portfolio renders `—` for Day P/L and Total Return instead of a fabricated `+$0.00 (0.00%)` sitting next to real figures.
- Portfolio renders guarded by a sequence token; dropped the duplicate `GET /api/v1/portfolio` on every My Agents tab switch.

**Tests** — 1197 passed / 34 skipped. New: legacy/claimed-agent reconciliation, failed-PATCH isolation, transfer auth/ownership/bounds, threaded over-allocation, `set_cash_available` on the Postgres tier, and node-run frontend checks. All mutation-verified (the threaded one only fails without the lock once the sleeve write is slowed — the GIL's 5ms switch interval otherwise hides the window).

**Not done:** `/allocate` and `/reclaim` still aren't called from any frontend code. Building that UI looked like scope beyond a review fix — say the word if you want it here rather than a follow-up.

⚠️ The `@pg_only` tier skips locally (no local Postgres), so the new Postgres tests need a CI log check, not just a green tick — that `skipif` fails open.

FlyM1ss:
Filed the unwired UI as #191 — that resolves the open question from my review comment: allocate/reclaim stay backend-only in this PR, and the panel wiring lands separately.

Issue #148, opened 2026-07-18
Investigate Nemotron’s Poor Leaderboard Performance


Allan-Feng:
Nemotron has been connected to the leaderboard through OpenRouter, but its performance curve is significantly worse than the other models.

The goal of this issue is to carefully debug the existing Nemotron run and identify the cause of the poor performance. Please inspect the complete path from the model response to the final portfolio curve, including:

- Whether the correct Nemotron model and configuration are being used;
- Whether the model responses follow the expected trading format;
- Whether responses are parsed and executed correctly;
- Whether errors, invalid outputs, or rejected actions are being converted into `HOLD`;
- Whether the poor curve is caused by the model’s actual trading decisions.

Students are encouraged to investigate independently, compare observations, test reasonable hypotheses, and propose small fixes. Focus on understanding this specific curve rather than building a new evaluation framework.

### Expected Outcome

Provide a short summary of:

1. What was checked;
2. What problem was found, if any;
3. Evidence supporting the conclusion;
4. The recommended fix or next experiment.

MrParamecium:
## What was checked

I checked the Nemotron model ID, OpenRouter reasoning configuration, response content blocks, JSON parsing, retry and fallback behavior, action execution, simulated fills, and the final portfolio curve.

## Problem found

OpenRouter's default `medium` reasoning setting maps to a 2048-token reasoning budget, while ATL limits the entire model response to 2000 output tokens.

In the failing run, Nemotron frequently consumed the full output budget on thinking blocks and returned no JSON text. Those requests were retried and eventually became rescue calls, fallback decisions, or effective HOLDs. Therefore, the original curve was not purely produced by Nemotron's trading decisions.

## Fix

I added an optional per-model `reasoning_effort` setting and configured only the Nemotron leaderboard entry with:

```json
"reasoning_effort": "none"
```

The setting is passed from the leaderboard config through `LLMAgentStrategy` to the OpenRouter client instance. Other OpenRouter, CommonStack, and Anthropic models keep their existing behavior.

Branch: `fix/issue-148-nemotron-reasoning`

## Evidence

A post-fix Alpaca DJIA-30 run over 2026-04-15 through 2026-05-15 produced:

- 161 decision steps
- 161 API calls
- 155 model-driven decisions
- 96.27% model-decision coverage
- 0 no-text responses
- 0 retries
- 0 request errors
- 134 simulated trades
[...]

PR #189, opened 2026-07-23
fix(leaderboard): make Nemotron runs reproducible

MrParamecium:
## Summary

- add an optional per-model trading `temperature` and pin Nemotron to `0`
- persist the effective model ID, integration, temperature, reasoning effort, output-token ceiling, capital, and date window on each LLM leaderboard run
- migrate legacy `equity_timeseries` data to a unique `(run_id, timestamp)` key so forced reruns replace points instead of appending duplicate curves

Refs #148.

## Corrected diagnosis

The historical website curve (`-0.22%`) was already run with `$100,000`, `LLM_MAX_OUTPUT_TOKENS=16000`, and reasoning disabled. It was not a 2k/medium-reasoning truncation run, so the earlier 2k reasoning explanation does not apply.

Three controlled runs per configuration over 2026-04-15 through 2026-05-15 produced:

| ceiling / reasoning / temperature | return, mean +/- sample std | mean model-decision coverage | mean invalid JSON |
|---|---:|---:|---:|
| 16k / off / provider default | `-1.514% +/- 0.529 pp` | 97.93% | 3.3 |
| 16k / off / 0 | `-0.065% +/- 0.325 pp` | 98.34% | 2.3 |
| 2k / off / 0 | `-0.002% +/- 0.203 pp` | 96.07% | 6.0 |

The default-temperature group had no request errors and high model-decision coverage, yet all three runs lost money. Its valid responses covered only about 42.7% of the input symbols, versus 74.7% at temperature 0. This supports the conclusion that the poor curves came mainly from the model's sampled trading output under an unpinned temperature, not from parser or execution corruption. The 2k ceiling does increase malformed-JSON effective HOLDs and leaves H6 coverage near the 95% boundary, but it does not explain the historical curve.

Temperature 0 reduces variance but does not guarantee bit-for-bit deterministic provider output, so leaderboard conclusions should use repeated runs rather than one draw.

## Tests

- focused temperature/metadata/migration tests: `95 passed`
- backend suite excluding one pre-existing vn.py test: `1197 passed, 29 skipped, 1 deselected`
- the deselected `test_canonical_simulation_runs_agent_and_both_baselines_offline` also fails unchanged on `origin/main@3af0457` because the rule-based simulation makes zero trades in that fixed window


FlyM1ss:
Reviewed and pushed fixes for the findings (`b988f24`, `30c5cd5`).

**Verified first** — the `MAX(id)` retention rule is empirically the right one. For both affected runs the surviving curve's last point matches the published summary row exactly:

| run | `agent_runs.final_equity` | surviving curve | dropped curve |
|---|---|---|---|
| Sonnet 4.6 | 100114.14 | **100114.14** ✅ | 101097.17 |
| Nemotron | 99779.94 | **99779.94** ✅ | 96770.04 |

So the migration reconciles the chart with the table rather than repointing a published entry at a curve nobody scored — worth stating in the PR body.

**Fixed**

1. **A transient `database is locked` hard-failed app boot.** `database.py` ends with a module-level `db = BacktestDatabase()`, so raising in the migration failed the *import* of `dashboard.backend.database`. Reproduced against a real exclusive lock: the neighbouring migrations degraded (`⚠️ Migration warning: database is locked`) and this one then raised `RuntimeError`. Now retried once with a fresh connection, then deferred with a warning; `IntegrityError` and the post-create verification failure still abort. Re-verified against the same real lock — startup survives and the DB is left untouched for the next boot.
2. **The committed seed `backtest.db` was not migrated**, so every local `uvicorn` run would silently rewrite a tracked binary. Migrated and committed. Delta proven with a line-by-line `.dump` diff: exactly 483 `equity_timeseries` INSERTs removed, one unique index added, nothing else. Re-opening it through the app is now byte-identical, with no WAL sidecar left behind.
3. **Force-refresh could still merge leftover points.** The unique key collapses timestamps that *repeat*; it can't remove points the new curve no longer produces. `insert_equity_points` now replaces the run's curve in one transaction (empty input stays a no-op; `replace=False` preserves the append path one test fixture uses).
4. **`temperature` accepted 0–2** but every gateway here is an Anthropic Messages client capped at 1.0 — narrowed to 0–1, and pinning it alongside a reasoning-enabling `reasoning_effort` is now rejected outright instead of becoming a per-step 400 → silent rule-based fallback.
5. Nits: `ensure_leaderboard_runs` records the same metadata snapshot as `deploy_model_run`; `temperature` typed `Optional[float]`. The plan/spec docs were amended where they now describe the old behaviour.

**Retracted from my earlier notes:** the Chinese plan docs missing a `.zh-CN.md` suffix is *not* a deviation — `plans/` already follows that convention (`2026-07-19-nemotron-reasoning-none-fix.md` is Chinese too); only `specs/` uses the suffix.

Full backend suite: **1208 passed, 30 skipped**.

Follow-ups filed: #194 (published entries' configs have drifted from the registry and carry no provenance — needs a re-run) and #195 (`_migrate_trades_schema` can never converge, so it re-runs on every boot; pre-existing, and it widens the very lock window fixed above).

PR #192, opened 2026-07-23
fix: address CodeQL baseline findings

FlyM1ss:
Response to the first full CodeQL scan (1079 alerts). Two commits:

**`ci:`** exclude `orchestration/` from the scan — imported research code, 771 of the 1079 alerts; its alerts auto-close on the next `main` scan.

**`fix:`** the findings that were real:
- **Dead endpoint** (#67): `POST /paper/start-session` omitted `insert_run`'s required `session_id` → every credentialed call failed with a swallowed `TypeError`.
- **SSRF** (#48–50): ticker symbols were interpolated raw into Alpaca URL paths; now allowlist-validated (`BRK.B`/`BF-B` style still accepted).
- **Leaderboard stored XSS** (#4, under-flagged): `entry.model`/`team_name` (user-registered names) rendered unescaped into `innerHTML`; templates extracted into pure renderers with full escaping, proven by node-executed round-trip tests.
- **Exception text in responses** (#12–21): `{"error": str(e)}` envelopes now `print()` server-side and return generic messages; source guard keeps the pattern out of `api/routers`.
- **`/images` traversal guard** (#46–47): only static route without the separator/containment check its siblings have.

Suite: 1190 passed / 30 skipped (baseline 1175/30; 15 new tests, all verified red first). The ~23 false-positive alerts are being dismissed via the API with per-alert reasons, not code changes.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -22,3 +22,9 @@ paths-ignore:
   # Vendored / generated JS is never our code to fix.
   - "**/node_modules"
   - "**/dist"
+  # Imported research code (NeurIPS 2025 FinAgent framework) — per CLAUDE.md a
+  # separate project, not wired into the shipping dashboard or deployed. Its
+  # first scan contributed 771 of 1079 alerts (including files that don't
+  # parse); scanning it buries the dashboard findings we act on. Excluding a
+  # path auto-closes its existing alerts on the next default-branch scan.
+  - orchestration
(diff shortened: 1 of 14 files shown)


FlyM1ss:
Merged; post-merge wrap-up:
- Main scan green; `orchestration/` exclusion auto-closed the imported-research alerts (repo went 1079 → 191 open, all quality-tier, **0 security-severity**).
- Dismissed 6 modeling FPs the scanner re-raised on the fixed code: #1102/#1104/#47 (`/images` — CodeQL treats `resolve()`/`is_file()`/`FileResponse` as sinks even behind the containment guard) and #48–50 (`quotes.py` — regex allowlist not modeled as a taint barrier). Guards are test-covered (`test_static_file_guards.py`, `test_quotes.py`).
- Remaining 191 quality alerts tracked in #193.

PR #209, opened 2026-07-24
feat: background baseline worker + finalize split

FlyM1ss:
T2 of `docs/superpowers/specs/2026-07-24-agent-scale-sustainability-design.md`.

## Summary
- Baseline generation moves off the finalize request into a lazily-started, single-consumer `baseline_worker` daemon. The final decision submit now returns in ms (was seconds-to-minutes holding `_step_lock`).
- Config-level dedup: a same-config wave runs **2** baseline backtests (buy-hold + DJIA), not 2N. Runs share one baseline pair via `_completed[(start,end,mode)]`.
- Async publish self-heals: `_finalize` returns a completed envelope with empty `baseline_run_ids`; the worker swaps in a new dict (GIL-atomic) and later polls / the DB pick it up.
- New env var: `BASELINE_QUEUE_MAX` (default 500). Overflow / failure degrade exactly like the old best-effort inline path (run saved, baselines absent).

## Dedup safety (spec Step 7 — verified at source)
Shared baseline rows are written under the *first* finalizer's `session_id`, so both consumer paths were confirmed non-session-scoped:
- `db.delete_run(run_id)` deletes only that run's own rows — never the runs its `baseline_*_run_id` columns point at.
- The compare view reads baselines through the public `/compare` endpoint (`db.get_run`/`db.get_equity_curve`, explicitly "not filtered by session").

No fallback to per-run baselines needed.

## Notes
- The plan's fixture-grep (`tests/*.py`, non-recursive) missed three canonical-wiring move-tests under `tests/backtesting/` and `tests/domain/backtesting/`; the `HourlyBacktester` consumer moved service→worker, so those assertions were redirected to `baseline_worker`. Corrected one now-stale finalize comment as well.

## Test plan
- [x] `test_baseline_worker.py`, `test_finalize_split.py` (new) pass
- [x] Contract pins pass unchanged: `test_run_lifecycle_unification.py`, `test_protocol_api.py`, `test_execution_backends.py`, `test_pr71_review_fixes.py`
- [x] Full backend suite: 1282 passed, 34 skipped

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -86,6 +86,10 @@ DATABASE_PATH=dashboard/storage/data/backtest.db
 # in memory for reuse across backtest runs (LRU beyond this). Default 4.
 # MARKET_DATA_CACHE_MAX_ENTRIES=4
 
+# Max queued background baseline-generation jobs (drops beyond this; the run
+# itself is unaffected). Default 500.
+# BASELINE_QUEUE_MAX=500
+
 # User accounts (optional). When set, signup/login/session data is stored in
 # this Postgres database instead of the local SQLite file above -- required
 # on hosts with an ephemeral/disk-less filesystem (e.g. Render free tier),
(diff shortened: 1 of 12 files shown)


FlyM1ss:
GHAS follow-up: this PR introduced 1 CodeQL quality alert — #1111 (empty-except in domain/backtesting/baseline_worker.py). **Fixed** in #213 (explanatory comment; behavior-preserving); auto-closes on the next `main` scan. No Dependabot/secret-scanning alerts.

PR #220, opened 2026-07-25
fix(app): drop duplicate home hero + fabricated result drivers

FlyM1ss:
Two low-risk friction cuts on the in-app Home, from Allan's UI-funnel review.

- **A — no more double-hero.** The in-app Home re-showed the marketing landing hero as pager screen 0, so landing "Get Started" → `/app?view=home` dumped the user on a *second* identical hero + "Get Started". Deleted that screen; Home now opens straight on the real dashboard. Removed the now-dead hero JS (`initLandingPlaygroundChat`, `initHomeGetStarted`).
- **B — no fabricated result data.** The backtest result panel rendered a "Performance Drivers" card of three hardcoded lines (identical every run, one negative line styled positive) plus two buttons with no JS behind them. Removed. Real, data-driven metrics untouched.

Purely subtractive (185 del / 1 ins on the frontend). Adds 4 source-level guard tests (`test_frontend_home_friction.py`) since the frontend has no JS harness. Full backend suite green (1307 passed / 35 skipped).

https://claude.ai/code/session_01VhHXBK3cGQkW9XMcuaxEJm

### DIFF
--- a/dashboard/backend/tests/integrations/test_frontend_home_friction.py
+++ b/dashboard/backend/tests/integrations/test_frontend_home_friction.py
@@ -1,42 +1,17 @@
-"""Guard the A+B navigation/honesty friction cleanup.
+"""Guard the backtest-result honesty friction cleanup.
 
 The frontend is vanilla JS with no test harness, so these are source-level guards
-(read the files as text) that run in CI and lock two fixes:
+(read the files as text) that run in CI and lock one fix:
 
-  A) The in-app Home no longer re-shows the marketing landing hero. Home opens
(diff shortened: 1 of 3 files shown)


Allan-Feng:
Pushed a follow-up (`08c85aa`):

- **Kept** the in-app Home hero (the second welcome screen). Please leave that as-is.
- **Kept** removing the fake "Performance Drivers" block and the two buttons that did nothing.

PR description still mentions dropping the hero — that part is outdated; only the result-panel cleanup remains.

PR #222, opened 2026-07-25
feat(landing): point Test section CTA at Start Free signup

Allan-Feng:
## Summary
- Replace the Test section **Race this agent ↓** button with the shared **Start Free** landing auth CTA
- Opens the on-landing signup modal (same as nav/hero/Talk/Race/footer) instead of scrolling to `#race`
- Rebuild the shipped frontend bundle

## Test plan
- [ ] Hard-refresh `/`
- [ ] Scroll to Test section → button label is **Start Free**
- [ ] Click it → landing signup modal opens (not jump to Race)
- [ ] Close modal → stay on landing


FlyM1ss:
Reviewed — the CTA change is correct and matches the `Race.tsx`/`Talk.tsx` pattern exactly. I pushed two commits; both are additive, neither touches your change.

**1. Merged `main` in.** The branch was based on `c7234ea`, so its green CI predated the bundle guard test (#226) and never ran it. Now it does.

**2. Verified the bundle, then added a guard for the gap it exposed.** I rebuilt `dashboard/landing` from this branch's source: it reproduces `index-CGH4sA_a.js` and `index-CqmHw0ZJ.css` **byte-for-byte**, so the committed bundle is a faithful build — no hand-edits, nothing stale.

But nothing enforced that. The existing checks prove the bundle *loads*, not that it's *current*, so a source edit that was never rebuilt would ship silently. Two checks added:

- one `data-landing-auth` per `landing/src` emitter (6 == 6 — the count your CTA moves to 6)
- every `lib/cta.ts` label present in the bundle, which catches copy drift the count can't see

Both compare only minifier-stable markers (string literals, quoted hyphenated keys), so they don't reintroduce the build-reproducibility flakiness #226 deliberately avoided. Mutation-verified: adding a 7th CTA or renaming the label without rebuilding each fails with the offending file named. Full suite 1562 passed / 0 failed.

Also browser-verified `/`: Test-section button reads **Start Free**, opens the signup modal, no jump to `#race`, `#race` still reachable from the navbar.

Unrelated, for visibility: #228 clears the one open Dependabot alert (`postcss`, dev-scope, in the landing lockfile).

PR #229, opened 2026-07-27
chore: dedupe TradingAgents internals, refresh SDK docs

FlyM1ss:
Follow-up to #224 (already merged), so this is a fresh branch off `main`.

**Refactor** — `_tradingagents_runner.py` had forked the SDK's own run lifecycle: re-declared `_FAILED_STATES` and `_STEP_AUTOHELD_CODES` (whose `runner.py` comment says the backend contract is locked by a test and "Do NOT drop either"), plus a copy of `_wait` that had already drifted. All three now come from `runner.py`. Also: duplicate ISO-date validator, an unreachable `missing_price` branch, per-record-per-step `date.fromisoformat`, stringly-typed `setattr/getattr` counters, and a hand-maintained export list that had silently lost three names.

**Docs** — the SDK quickstart's example catalogue (linked from the PyPI README) never listed the TradingAgents bridge; the "no third-party dependencies" line predates #224's `tzdata` on Windows; `external_agents.rst` claimed a 30s decision timeout while contradicting itself further down (server default is 60s); and `timeout_holds` was missing from both the protocol-doc metrics example and `models.py`'s docstring, which is the SDK's only schema reference for `result.metrics`.

Public API unchanged. Backend suite 1562 passed / 35 skipped; SDK suite 117 passed (5 new).


FlyM1ss:
Follow-ups from the cleanup pass, filed rather than left in this thread — each re-verified against `116874e` before filing:

- #230 — `AgentRunner`'s `decide()` seam can't host the TradingAgents loop, so `run_backtest` stays forked. Gated on the SDK 0.2.0 / v2 migration.
- #231 — `run_from_args` is library orchestration living in `dashboard/examples/`.
- #232 — the CLI's default artifact-path branch is never exercised (low severity).

One finding from an earlier scan was **not** filed because it's a false positive: `docs/source/lab/external_agents.rst` was reported as missing the `429 too_many_active_runs_global` error. That page documents the legacy `/api/v1/backtest/*` routes; the global cap is enforced only in `domain/runs/service.py:446` (protocol `/api/v1/runs`) and `api/v2/runs.py:358`. Documenting it there would tell readers to handle an error their endpoint cannot return.

PR #233, opened 2026-07-27
feat(sdk): run vn.py CTA strategies on ATL

MrParamecium:
## Summary

- expose the current OHLCV bar in protocol observations for external strategy runtimes
- add an optional vn.py CTA integration for trusted local `CtaTemplate` strategies
- adapt vn.py `buy` and `sell` signals into ATL orders with T+1 execution semantics
- persist deterministic audit artifacts with strategy inputs, captured orders, ATL responses, and diagnostics
- add a runnable double-moving-average example, an English quickstart, and English design and implementation documents

## Validation

- `171 passed` in `packaging/agentictrading/tests`
- `1571 passed, 34 skipped` in `dashboard/backend/tests`
- `222 passed` across the SDK and directly affected protocol/backtest tests after the documentation cleanup
- deterministic offline smoke produced a complete `buy -> sell` loop
- real Alpaca hourly AAPL smoke completed 105 steps with no adapter, timeout, or fatal-data errors; the strategy emitted three one-share buy orders, all correctly rejected by the ATL `max_position_weight=0.25` constraint because AAPL was above $250 in a $1,000 account
- no Chinese text or Chinese filenames in the PR diff or commit history
- `git diff --check origin/main...HEAD`

## Scope

This first version runs trusted local, bar-driven vn.py `CtaTemplate` strategies. It does not execute untrusted uploaded Python, connect a live broker, or reproduce the full vn.py CTA engine. Hosted smoke runs require this branch to be deployed first because the current production protocol does not yet include `market.bars`.

### DIFF
--- a/dashboard/backend/domain/backtesting/external_run_service.py
+++ b/dashboard/backend/domain/backtesting/external_run_service.py
@@ -312,7 +312,7 @@ class ExternalBacktestSession:
         for symbol in symbols:
             signal = signals[symbol]
             snapshot["top_signals"][symbol] = {
-                "price": float(signal.get("price") or 0),
+                "price": float(signal.get("price") if pd.notna(signal.get("price")) else 0),
                 "rsi": float(signal.get("rsi") if pd.notna(signal.get("rsi")) else 50),
                 "macd": float(signal.get("macd") if pd.notna(signal.get("macd")) else 0),
                 "macd_signal": float(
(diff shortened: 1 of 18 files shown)


FlyM1ss:
Pushed fixes for the review findings above (3 commits, all CI green: Backend tests, Packaging tests, Analyze python/js, CodeQL).

**Confirmed bug fixed** — `protocol_bars()` now drops a symbol that fails its OHLCV sanity check instead of raising, so a malformed bar can no longer permanently 500 a run's `/steps/next` poll loop. Also NaN-guarded `top_signals`' `price` field (same request path, same root cause — a NaN close was still leaking into JSON serialization through a second field even after the bars fix) and fetched market data once per step view instead of twice. New regression tests cover the drop-not-raise path at both the unit and HTTP-integration level.

**Other findings addressed:**
- Fill/rejection matching now prefers the closest `requested_quantity` match among same-symbol/side candidates, not just the first one — reduces (doesn't eliminate, since `orders-v1` has no client order id) misattribution risk when a strategy emits multiple same-side orders in one bar.
- Deduped the OHLCV validation logic shared by the adapter and runtime (same package) into one `validate_ohlcv_values()` helper. Left the backend's copy separate — different package (can't share code between the stdlib-only SDK and the FastAPI backend) and different failure semantics (drop vs. reject).
- Widened credential redaction from `key=value` only to also cover `key: value` and quoted/JSON-style shapes, and softened the doc's absolute "never stores the API key" claim to describe what's actually guaranteed vs. best-effort.
- Fixed the mislabeled "Chinese vn.py CTA guide" link (the doc is English).
- `VnpyCtaRuntime.drain_captured_orders()` is now a public method so the adapter's exception-recovery path doesn't reach into `runtime.engine` directly.

Checked GHAS: no code-scanning alerts open on this branch.

FlyM1ss:
Fixed one more: GHAS flagged `py/unused-import` (alert #1147) on `AtlCtaEngine` in the test file's import block — pre-existing in the original commit, unrelated to the fixes above. Pushed in `9785201`. All CI green, 0 open code-scanning alerts on this branch, mergeStateStatus CLEAN.

PR #235, opened 2026-07-27
fix: keep default Foundation agent after signup + land on Home

Allan-Feng:
## Summary
- Signed-in `list_agents` now also returns unclaimed browser-owned agents, so a guest-provisioned **My Foundation Agent** stays visible after signup (and claim) instead of disappearing while logout still showed it.
- Provision guard is user-scoped; boot claims before the first My Agents paint; concurrent provision waits instead of skipping.
- New account signup (landing + in-app) navigates to **Home** instead of restoring a leftover My Agents `nav-state`.

## Test plan
- [ ] Logged out: open My Agents → see My Foundation Agent
- [ ] Register a new account on the same browser → My Agents still shows that agent (or a new default)
- [ ] Fresh landing signup → lands on `/app?view=home`, not My Agents
- [ ] In-app signup while on My Agents → switches to Home
- [ ] Returning login still restores last tab
- [ ] `pytest dashboard/backend/tests/domain/agents/test_repository_move.py dashboard/backend/tests/test_agents_api.py -q`

### DIFF
--- a/dashboard/backend/domain/agents/repository.py
+++ b/dashboard/backend/domain/agents/repository.py
@@ -321,6 +321,10 @@ class AgentStore:
             )
 
         conn.close()
+        # Each _add_rows call is independently sorted, but the groups are
+        # appended query-by-query, so a recent unclaimed browser agent could
+        # otherwise land after an older owned agent. Re-sort the union once.
+        rows.sort(key=lambda row: row["created_at"], reverse=True)
         return [_public_agent(row) for row in rows]
 
     def list_builtin_agents(self) -> List[Dict[str, Any]]:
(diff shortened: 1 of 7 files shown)


FlyM1ss:
Pushed a follow-up commit addressing review findings:
- Global `created_at` ordering across owned + unclaimed-browser rows (both SQLite and Postgres twins), instead of two independently-sorted groups.
- `@pg_only` test coverage for the unclaimed-browser-session branch in the Postgres twin (previously untested, but that's what backs prod via `CONTENT_DATABASE_URL`).
- The onboarding guard no longer stamps "done" against a builtin only visible via the unclaimed fallback — only once it's actually owned by the signed-in user.
- Cross-referenced the duplicated `nav-state` key between `index.html` and `app.js`.

All backend tests green locally (`188 passed, 19 skipped` on agent-scoped tests; full suite green modulo one pre-existing order-dependent flake in `test_ifind_ashare_engine.py`, unrelated to this change).

PR #236, opened 2026-07-27
fix: default CommonStack to DeepSeek (Anthropic stub greeting)

Allan-Feng:
@/tmp/claude-1000/-mnt-d-github-agent-trading-lab/bb1c3697-ed3a-448e-b88f-b1caa1aff76a/scratchpad/pr236_body.md

### DIFF
--- a/dashboard/backend/domain/chat/service.py
+++ b/dashboard/backend/domain/chat/service.py
@@ -4,7 +4,7 @@ import os
 from collections import defaultdict
 from typing import Any
 
-from anthropic import AsyncAnthropic
+from anthropic import APIError, AsyncAnthropic
 from dotenv import load_dotenv
 
 from dashboard.backend.infrastructure.llm.backtest_harness import (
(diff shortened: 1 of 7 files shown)


FlyM1ss:
Pushed 762fa63 addressing review feedback:
- Fallback now also triggers on a hard provider error (`anthropic.APIError`), not just a detected stub reply, in both `synthesize_strategy_prompt` and `chat_with_agent`.
- `/ask` (`chat_with_agent`) gets the same stub-detection + fallback as `/strategy`.
- The stub-exhaustion error only mentions CommonStack/`CHAT_MODEL` when a CommonStack key is actually set.
- 9 new tests cover all of the above; full backend suite still green (1726 passed, 35 skipped).

Checked the two automatable test-plan boxes (ran the pytest command myself; the CommonStack-default assertion is covered by `test_providers.py`). Left the two live-Discord boxes unchecked since I can't drive an actual Discord session here — a human should still click through those before merge, though the automated tests now exercise the same code paths (including a forced-stub scenario).

Issue #187, opened 2026-07-23
feat: password reset via email (Profile Phase 2) — blocked on email provider


FlyM1ss:
Phase 2 of the profile work started in #171. Designed but not started, because it cannot be built without an operator step.

Spec: `docs/superpowers/specs/2026-07-22-profile-system-enhancements-design.md` (§ "Phase 2 — Password reset via email") — `forgot-password` + `reset-password`, single-use expiring tokens, cooldown, and enumeration-safe responses (identical reply whether or not the address exists).

**Blocked on:** an email provider account (Resend in the spec) and three env vars set in the Render dashboard before the code merges, same as the `CONTENT_DATABASE_URL` rule in CLAUDE.md:

- `RESEND_API_KEY`
- `RESET_EMAIL_FROM`
- `PUBLIC_APP_URL`

Until then a user who forgets their password has no self-service path. Not urgent at the current account count; filing so the ops prerequisite is visible somewhere other than a design doc.

FlyM1ss:
The provider blocker here mostly dissolves once #237 lands, but the env-var list in this issue is now wrong — **Brevo** was chosen, not Resend.

What #237 already builds:

- `dashboard/backend/infrastructure/email/sender.py` — a provider-neutral `send_email(to, subject, text_body) -> bool`. Callers get a bool, never an exception, so an outage cannot 500 a route; every failure path `print()`s an ERROR first (`logger.*` is invisible under the deployed config) and the caller is expected to return 503 rather than a cheerful `ok`. Config is read at call time, so tests can set and unset freely.
- `dashboard/backend/verification_codes.py` — `generate_code()` / `hash_code()`, 6 characters from a 31-symbol ambiguity-free alphabet, SHA-256 hashed. This module was deliberately split out of `api/auth.py` **for this issue** to be its second consumer.
- The rate-limiting shape a reset flow needs, already exercised: cooldown that survives cancel, rolling daily cap, `Retry-After` carrying the real remaining wait.

Corrected prerequisites — two vars, not three, and both are already in `.env.example`:

- `BREVO_API_KEY`
- `ACCOUNT_EMAIL_FROM` (plus optional `ACCOUNT_EMAIL_FROM_NAME`)
- ~~`RESEND_API_KEY`~~ / ~~`RESET_EMAIL_FROM`~~ — superseded.
[...]

PR #237, opened 2026-07-27
feat: account identity editing

### LINKED ISSUE #185: fix: test_auth.py client fixture patches a store the routes never read
Pre-existing, found while reviewing #171 (not introduced there).

`tests/test_auth.py`'s `client` fixture builds a temp `UserStore` and patches it in:

```python
monkeypatch.setattr(users_module, "user_store", temp_user_store)
```

`api/auth.py` binds `user_store` at import time (`from dashboard.backend.users import ... user_store`), so it keeps its own reference. Rebinding the attribute on `users` afterwards does not reach it — `get_current_user` and every auth route still use the module singleton, and the per-test temp store is never exercised.

Verified inside a test run:

```
users_module.user_store is temp_store -> True
auth_module.user_store  is temp_store -> False   (points at conftest's session-wide temp DB)
```

Impact today is nil: emails are distinct across the file and conftest keeps writes off the committed seed DB. It is a trap for the next test — reusing an email, or asserting "no users exist", fails for a reason the fixture actively hides. `test_change_password_revocation_failure_still_succeeds` already works around it by patching at the class level.

Fix is the pattern already used by `pg_client` in `tests/test_users_postgres.py` and by `tests/test_discord_linking.py` — patch both bindings:

```python
monkeypatch.setattr(users_module, "user_store", temp_user_store)
monkeypatch.setattr(auth_module, "user_store", temp_user_store)
```

`api/routers/discord.py` holds a third import-time binding; worth patching too for any test that touches those routes.

FlyM1ss:
`BREVO_API_KEY` and `ACCOUNT_EMAIL_FROM` are now set on the Render service (2026-07-28; 15 → 17 vars, no existing var touched). An env-var write does not redeploy here, so they take effect on the deploy this merge fires — code and credentials land in the same process start.

Display-name editing; email change behind the current password plus a code to the **old** address (proves account control) then a code to the **new** one (proves reachability); and the account-page Log out button moved to the bottom in destructive red.

Closes #185 — `/api/auth` routes now resolve `user_store` at call time, at both import-time binding sites.

## Rate limiting

Three windows, checked widest-first so the 429 reports the longest wait that actually applies:

| Window | Keyed on | Purpose |
|---|---|---|
| **7 days** | last **completed** change (`used_at`) | Product policy — an address is a contact detail, not something to churn. |
| **3 per rolling 24h** | request `created_at` | Bounds the shared provider quota. |
| 60 seconds | newest request | Unchanged. |

Keying the 7-day limit on a *completed* change rather than a *request* is the whole point: widening the existing cooldown to a week would cost a user seven days for one mistyped address. The daily cap is what bounds abuse, because the drain path — request, read your own code, verify, mail an arbitrary third party, cancel, repeat — never completes. It takes one account from draining the shared 300/day quota in ~2.5h to needing ~50 accounts. The residual (quota is platform-wide, signup is unlimited) is #243.

Both limits read history off `email_change_requests`, so **that table is now append-only**: `create` supersedes prior rows instead of `DELETE`ing them, and `cancel` is scoped to still-active rows so a completed change is not relabelled as cancelled. A `DELETE` there silently disables both limits — mutation-tested, it fails 6 tests across the store and route layers.

`Retry-After` carries the remaining wait, not the window width; at seven days those differ enough to matter.

## Test plan
- [x] Full suite **1799 passed / 41 skipped / 0 failed** — `main` baseline is 1715 / 35 / 0, so additions only
- [x] Mutation-tested: neutering the 7-day check, the daily cap, the supersede-not-delete, and the cancel scoping each fails tests that pass otherwise
- [x] `node --check dashboard/frontend/app.js` (the suite reads `app.js` as text and cannot catch a syntax error)
- [x] Seed DB byte-identical to `main`; no commit on this branch touches it
- [x] `@pg_only` tier verified against a real `postgres:18-alpine` container on CI

## Notable
- Both mail steps **send before persisting**, so a delivery failure never advances state past a message that was never delivered.
- The plan's 32-symbol code alphabet was unsatisfiable against its own test (it still contained `L`); shipped as 31 symbols with all five ambiguous characters removed. Confirmed as intended.
- CodeQL: 1152 (`py/import-and-import-from`) fixed; 1151 (`py/cyclic-import`, deferred function-scope import is the cycle-break) and 1150 (`py/missing-equals` on a list-subclass test double whose list-equality is load-bearing) dismissed with reasons. 1153 (`py/side-effect-in-assert`) caught a real regression I had introduced and is fixed.

## Follow-ups filed
- #239 — the `accounts.rst` manual page (draft, gated on this PR).
- #240 — give accounts a durable identifier before entitlements are keyed on email.
- #241 — nothing in CI builds `docs/source`, so broken RST ships to RTD.
- #242 — Robinhood link/unlink logs are invisible in prod (pre-existing, from #227).
- #243 — the transactional email quota is shared platform-wide with no global ceiling.
- #187 updated: its Resend env-var list is superseded by Brevo, and its blocker is largely dissolved.


FlyM1ss:
### CI + code scanning

All checks green. **`@pg_only` tier verified**: CI stood up a real `postgres:18-alpine` container and ran `test_users_postgres.py` — 17 tests, **zero skips**. That tier skips locally (`TEST_POSTGRES_URL` unset), so this is the first execution of the Postgres twin's SQL anywhere.

CI: 1800 passed / 0 failed. Local: 1782 passed / 40 skipped / 0 failed (`main` baseline 1715 / 35 / 0 — additions only).

### CodeQL: 0 security alerts, 5 quality alerts — 2 fixed, 3 left for your call

**Fixed** (`f8c7878`) — `py/side-effect-in-assert` ×2, both errors:
`assert client.post(...).status_code == 200` put a state-mutating request *inside* the assert expression. `python -O` strips asserts, so the request would never fire and later assertions would silently test state that was never created. Hoisted into named locals (also gives better failure messages). CodeQL sampled 2; I found and fixed **11** across 7 tests. Read-only `assert client.get(...)` forms were left alone deliberately — those calls mutate nothing, so stripping the assertion is exactly what `-O` means. Suite counts identical before and after, confirming a pure hoist.

**Not dismissed — analysis below, your call.** I didn't want to autonomously dismiss alerts on a shared repo:

- **#1150 `py/missing-equals`** (warning) — false positive, provably. `_Outbox(list)` adds `ok` without overriding `__eq__`, but three tests do `assert sent_emails == []`, relying on the *inherited* list `__eq__`. Overriding it to include `ok` would break exactly those assertions. The missing method is load-bearing by its absence.
- **#1151 `py/cyclic-import`** (note) — pre-existing. `origin/main`'s `users_postgres.py:18` already imported from `users`; this branch only extended the name list. The reuse is deliberate: both twins share the literal same `public_user` / `_utcnow_iso` objects, which removes format drift structurally rather than by convention. (Same rule was dismissed as by-design in #177.)
- **#1152 `py/import-and-import-from`** (note) — pre-existing style in `test_users_postgres.py`; it mixed both import forms at lines 40/60/177 and 203/215/278 before this branch.

FlyM1ss:
Review follow-ups landed:

- e4ab99c — stage-two mail gets its own copy (the recipient may have no account, so no "your account" / "change your password"); code hashes compare via `hmac.compare_digest`; the daily-cap test now gives rows genuinely distinct ages (as its comment claimed) and pins `Retry-After` to the oldest-row window.
- 72035b9 — the email-change form resets on sign-out (its stage closure previously survived logout→login in the same tab), with a structural guard test.

GHAS: 0 open code-scanning alerts on this PR — 4 fixed on-branch, 3 dismissed with documented rationale (incl. the SHA-256 code-hash threat model).

Issue #140, opened 2026-07-17
Agent run history still evaporates on every deploy


FlyM1ss:
#134 makes agents, agent versions, and strategies durable via `CONTENT_DATABASE_URL`. It deliberately leaves **run history** (`agent_runs` and friends) in the ephemeral SQLite at `DATABASE_PATH` — spec Decision 1, phase-2 deferred.

Consequence, worth stating plainly now that the rest is durable: the live Render service is free-tier with no persistent disk, so `DATABASE_PATH` still resets to the git-committed seed `backtest.db` on **every deploy**, and merging to `main` **auto-deploys**. So:

- Every backtest run and its decision log dies on each deploy.
- The nightly LLM leaderboard entries written to `agent_runs` (the daily leaderboard work) still evaporate — an agent will survive a deploy while its entire run history does not.
- `defaults.json` pins seed `agent_runs` IDs, so the seed's own runs come back; anything created after the last deploy does not.

This is a **scope question, not a defect**: the split is intentional and documented. Raising it because "agents are persistent now" is easy to over-read, and the daily-leaderboard feature has a direct stake in it.

Options, roughly in cost order: move `agent_runs` to `CONTENT_DATABASE_URL` too (same twin pattern, but it's the largest and hottest table); or put a persistent disk on the Render service (paid tier) and let SQLite keep everything; or accept and document the split.

FlyM1ss:
**Decision (2026-07-18): accept + document the split for now — keeping this issue open as the standing tracker.**

Run history (`agent_runs` + decision logs) stays in ephemeral SQLite; nightly leaderboard entries continue to reset on deploy. This is the documented phase-2 deferral (spec Decision 1; CLAUDE.md "Prod deploy reality"), and we're consciously keeping it rather than acting now.

Deliberately **not closing** this — it's the tracker for the deferred durable-run-history work, so the intent lives in an issue rather than a merged PR comment.

Upgrade path when durable run/leaderboard history becomes a priority, in cost order:
1. **`PostgresRunStore` twin** under `CONTENT_DATABASE_URL` (same pattern as agents/versions/strategies). Caveat: `agent_runs` is the hottest/largest table — every backtest step writes decision rows — so unlike the low-write config-table twins, this needs its own design pass on Neon free-tier connection/compute limits, not a copy of the pattern.
[...]

PR #262, opened 2026-07-29
feat: durable run history

FlyM1ss:
**✅ The backfill has run — the merge gate is discharged.** It was executed against prod on 2026-07-30 *before* merging rather than after, which is strictly safer: `main` contains no reference to `AGENT_RUNS_DATABASE_URL` anywhere (`database.py:998` is still the bare `db = BacktestDatabase()`), so the deployed app could not read the runs database yet and pre-populating it was invisible to prod. The first deploy that *can* read Postgres will find it already populated, so the blank-`/runs` window never opens at all. The leaderboard was never affected either way (its `lb_*` tables stay on SQLite).

Moves the five cold run-history tables (`agent_runs`, `equity_timeseries`, `trades`, `backtest_decisions`, `run_manifest`) to a dedicated Neon project selected by `AGENT_RUNS_DATABASE_URL`, behind a signature-identical `PostgresBacktestDatabase` twin. Unset var ⇒ today's behavior unchanged.

The hot per-step table `idempotency_keys` stays on local SQLite via delegation, so no agent request gains a network round-trip.

## ⚠️ This PR removes a route

`DELETE /admin/clear` is **gone**, not authenticated. It called `db.clear_all()` with no auth at all — the session middleware only checks that `X-Session-Id` parses as a UUID — and once run history is durable, one anonymous request destroys the only copy. Zero callers, and no admin tier exists to gate it with. Both route freezes in `test_app_composition.py` updated; `db.clear_all()` itself stays for tests and `backtest_hourly_agent.py --clear`.

## Pre-merge verification (done)

- **Backfill executed against prod and independently verified.** 17 runs / 2102 equity points / 0 trades / 0 decisions / 0 manifests migrated into the dedicated `ATL-runs-main` Neon project, which was genuinely empty beforehand (0 public tables, PostgreSQL 18.4 — no half-migrated state). Verified by comparing Postgres back against the SQLite seed rather than trusting the script's own report: identical `run_id` sets, all three `config/defaults.json` runs present, `created_at` preserved on all 17 (the field whose reset mis-pairs whole baseline curves in `app.js`), per-run equity counts matching across all 17 runs, and identical metrics on the default agent run.
- **Idempotency exercised on real data, at the only moment failure was free.** A second full backfill left every count unchanged with no duplication — worth doing before the merge, since `main` cannot read this database yet, so a re-run that wrongly doubled `equity_timeseries` would have been harmless and fixable rather than a prod incident.
- **The committed seed DB was never written to.** `sha256` unchanged and `git diff` empty across both real runs, with the `-wal` sidecar still 0 bytes — the script redirects `DATABASE_PATH` to a scratch file before importing any backend module and reads from a copy, so the usual "importing a store module runs lazy `ALTER`s against the committed prod DB" trap does not fire.
- **`@pg_only` tier executed.** `1996 passed, 7 skipped` — the whole-run skip count drops from ~64 locally to 7 on CI, which is how you tell the 20 live-Postgres tests actually ran rather than silently skipping (`skipif` on this tier fails *open*).
- **Prod env verified via the Render API**, not from memory: `AGENT_RUNS_DATABASE_URL` is present on the web service, pooled (`-pooler`), and byte-identical to the value used in verification (SHA-256 compared; no credential printed).
- **Topology invariant holds in prod.** Run history is a genuinely separate database from `CONTENT_DATABASE_URL`/`USERS_DATABASE_URL` (which are co-located with each other, as intended). This is what keeps the largest, hottest tables' growth off the auth-critical database.
- **GHAS: no open alerts.** CodeQL is clean on `refs/pull/262/merge` (the branch ref reports a false all-clear); the analysis is confirmed present on this head, and the 2 remaining `results` are `py/cyclic-import` dismissals that match 6 existing store-twin precedents. Dependabot: 0. Secret scanning is not readable from a non-admin account, so the diff was scanned by hand — every connection string in the branch is a fake, and one real prod endpoint id that had leaked into the spec prose was truncated.

## Notes for reviewers

- `database.py`'s three lazy-migration loops now emit **literal** SQL instead of f-strings. Zero behavioral change — but the twin-parity guard reads DDL from source text, and interpolated SQL was invisible to it. This took the guard from seeing 4 of 16 lazily-added columns to all 22.
- `test_database_lazy_migrations.py` compares declared **type, notnull and default** (not just column names) and cross-checks the ALTER path against `CREATE TABLE`. That cross-check found a pre-existing divergence: `trades.quantity`/`side`/`value` are NOT NULL when created fresh but nullable on the migrated path, because SQLite rejects `ADD COLUMN ... NOT NULL` without a default. Latent (every writer coerces first) and older than this work — pinned in an explicit list that fails both if a fourth appears and if this one is silently resolved, not fixed here.
- `test_store_twin_parity.py` gains a named, commented exemption for `idempotency_keys` (deliberately absent from Postgres) plus an anti-rot assert, and a parser fix for unspaced `UNIQUE(`.
- Timestamp normalisation (`as_timestamp_text`) is applied to **both** stores. `insert_equity_point(s)`/`insert_decisions` passed a `datetime` straight through: SQLite absorbed it via `sqlite3`'s deprecated default adapter (space separator), Postgres rejected it. Fixing only the twin would trade a 500 for a silent cross-store text divergence that inverts when Python drops that adapter.
- The backfill is idempotent **and** now re-runnable to completion: per-run failure isolation (previously one bad legacy row aborted `main()`, and every rerun re-failed on it, so runs after it never migrated), a per-run `conn.transaction()` block in `_restore_created_at` (it shares one connection, where a failed statement otherwise poisons every later UPDATE), and a `pg_advisory_lock` held across the write phase — the trades/decisions skip-logic is a read-then-write across two pooled checkouts, and neither table has a unique key to collapse doubled rows. The lock connection commits after acquiring so it does not sit `idle in transaction`, which Neon would eventually kill.
- The read phase is guarded too: `SourceData` reads every manifest up front, and `get_run_manifest` is the one reader with no JSON guard, so a single bad `manifest_json` aborted the run before any write — the one place no rerun could help. Both failure modes (raises, and the `null` literal that decodes to `None` without raising) are now reported and counted in their own accounting bucket.
- `db_pool` grows a public `close_all_pools()` (`_reset_for_tests()` now delegates to it, so the conftest teardown is unchanged) and the backfill script calls it from `__main__`. Without it a *successful* run printed four `couldn't stop thread 'pool-1-worker-N' within 5.0 seconds` warnings after `Backfill complete.` — exit 0 and no data affected, but it read as a partial failure. Closing from `__main__` rather than `main()` keeps the in-process tests (which call `main()` directly) free of process-wide teardown, and the helper reads `sys.modules` instead of importing so `--dry-run` still works with no psycopg installed. Re-run against the real target to confirm: the run now ends on `Backfill complete.`, and re-verification against the SQLite seed still passes all six checks.
- `database_postgres.py`'s docstrings now enumerate **all six** twin divergences. The previous "exactly two named divergences" and `insert_run`'s "nothing reads that reset" were both wrong — eight call sites order by `created_at`, and `app.js` pairs baselines to a run by a `created_at` *range*, so resetting it mis-associates whole curves rather than merely reordering a list.

## Still to do (not in this PR)

- Follow-ups from the spec's pass-2 audit: #264 filed (`scripts/backtest.py:397` `insert_run` missing the required `session_id`, verified live on `main` @ 9d46dc7). Hot-half durability (`idempotency_keys`, `protocol_runs`, `protocol_steps`) is **already** tracked by #140 — no second issue was filed; #140 should be rescoped to the hot half when this merges, since this PR resolves its cold half.
- #266 (separate, pre-existing on `main`) swaps the sample host in `test_postgres_url_guard.py` for the `ep-x` placeholder the rest of the suite uses. Independent of this PR — a `git merge-tree` check confirms the two do not conflict, and either can merge first.


FlyM1ss:
GHAS status: **0 open alerts** on `refs/pull/262/merge` (4 at `e2d9bdf`'s parent; Dependabot 0; secret scanning not enabled).

- `py/import-and-import-from` ×2 (`test_backtest_db_postgres.py`) — **fixed** in e2d9bdf. The dispatch tests need the module object for `monkeypatch.setattr`, so the module-alias form is the one that had to survive; the three `from ... import PostgresBacktestDatabase` sites now match it.
- `py/cyclic-import` ×2 (`database.py:1034` / `database_postgres.py:22`) — **dismissed, won't fix**. Store-twin pattern: the factory's function-scope import *is* the cycle-break. Same disposition as the four prior identical dismissals (#1151 users, #1076/#1092/#1100 portfolios).

Note the rollup `CodeQL` check passes even with alerts present, and the branch ref reports 0 while the merge ref had 4 — query `?ref=refs/pull/262/merge`.

PR #265, opened 2026-07-30
fix(ui): make dark-theme model select options readable

Allan-Feng:
## Summary
- Fix Agent Configure Model dropdown contrast: native `<select>` menus were light-backed with inherited light text
- Apply the same dark `color-scheme` / option styling to Create Agent and shared `control-select` pickers
- Bump `styles.css` cache-bust to `v=69`

## Test plan
- [ ] Open an agent Configure page → expand Model dropdown → all options are readable (not light-on-white)
- [ ] Create Built-in Agent modal → Model dropdown is readable
- [ ] Hard-refresh to pick up `styles.css?v=69`

### DIFF
--- a/dashboard/backend/tests/test_frontend_account_page.py
+++ b/dashboard/backend/tests/test_frontend_account_page.py
@@ -134,5 +134,5 @@ def test_cache_bust_versions_were_bumped():
     styles_version = int(re.search(r"styles\.css\?v=(\d+)", html).group(1))
     app_version = int(re.search(r"app\.js\?v=(\d+)", html).group(1))
 
-    assert styles_version >= 68
+    assert styles_version >= 69
     assert app_version >= 51
--- a/dashboard/backend/tests/test_frontend_select_contrast.py
+++ b/dashboard/backend/tests/test_frontend_select_contrast.py
@@ -1,0 +1,131 @@
+"""Guards for dark-theme readability of native form controls.
+
+A ``<select>`` popup is drawn by the browser, not by us: the list panel is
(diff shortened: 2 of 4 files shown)


FlyM1ss:
Reviewed and pushed two commits to this branch (thanks for `maintainerCanModify`).

**Heads-up — the `color-scheme: dark` landed on the shared `.date-input, .control-select` rule**, so it also changed how the UA draws the date field's calendar glyph: from dark to light. `filter: invert(0.8)` existed to flip that glyph *light*, so it then ran backwards and made the `#startDate`/`#endDate` picker buttons near-black on `--bg-input`. Verified with a Playwright render before/after; invert removed.

Also in the two commits:
- `.filter-select` gets the same treatment — it was the one shared picker class still missing it (`#tradingLogFilter`, `#backtestRunSelect`).
- `test_frontend_select_contrast.py`: 4 source-level guards (each mutation-tested), incl. one that fails if a new `<select>` ships without a dark-scheme rule.
- Dropped the dead `class="auth-input"` from `#builtinAgentModel` — no such rule exists; it is styled by `.auth-field select`.

CI + CodeQL green on `606e09d`; 0 alerts introduced. The contrast fix itself is sound — nothing else needed from my side.

FlyM1ss:
Merged as 5fc29f6. Filed #268 for the follow-up: consolidating the 4 per-component `color-scheme: dark` declarations onto `:root`, which would also cover the 3 native controls (`.agent-capital-input` spinners, the live-trading checkbox) that no rule reaches today. The guard added here only covers `<select>`.

Issue #225, opened 2026-07-26
chore(landing): fold the auth modal into the React source


FlyM1ss:
`dashboard/frontend/index.html` is a **hand-patched** build artifact: the Vite template (`dashboard/landing/index.html`) is 26 lines, the shipped file is ~400. The extra ~370 lines — auth-gate script, `#landingAuthModal`, `<style id="landing-auth-patch">`, and the delegated `[data-landing-auth]` click handler — cannot be produced by `vite build` and must be re-applied by hand on every bundle refresh (`dashboard/landing/README.md` documents the recipe).

Six React components now emit `data-landing-auth`, and their only consumer is that hand-written script in a different, generated file. The obvious way to refresh the bundle — copy `dist/index.html` over the shipped one — deletes the modal and turns all six CTAs into buttons that do nothing, with no console error.

The README already names the fix: *"folding the auth modal + gate into the React source would remove even this remnant, making the build output exactly the shipped page."*

Scope:
- Port the modal + auth-gate into `dashboard/landing/src` so `vite build` output ships verbatim.
- Drop the now-dead `'Get Started'` label fallback in the delegated handler (no button renders that label since all CTAs moved to the shared `PRIMARY_LANDING_CTA`).
- Then `test_frontend_bundle_integrity.py::test_hand_written_auth_layer_survives_a_bundle_refresh` can be retired.

Context: raised reviewing #222, which made Test the sixth component to depend on this. Pre-existing, not caused by that PR. Interim guard landed in `dashboard/backend/tests/test_frontend_bundle_integrity.py`.

PR #276, opened 2026-08-01
feat(landing): below-fold value proposition

FlyM1ss:
## Summary

A tester could not tell what the platform was for without clicking in and exploring. The narrative sections (Talk/Test/Race) each describe an act but never state the problem being solved or who it is for.

- **`WhyCare` band** between Hero and Talk — states the problem first ("You have an idea about the market. *Testing it properly is the expensive part.*"), then the three acts, then the three capabilities the landing never mentioned: model choice, the Python SDK / API, and Discord. Reuses the `01/02/03` mono-label pattern so it reads as one system rather than a bolted-on marketing block. Hero is untouched.
- **`01 — Talk` reframed** away from Discord-first. On-site plain-English authoring has existed since the agent editor shipped (`app.html:972`), and "join a Discord server" is exactly the friction this audience will not clear. Discord becomes step 3, not the way in. Structure, the `DiscordMock` visual and the `#talk` anchor all stay.
- **Dead footer links fixed.** `Terms`/`Privacy`/`Documentation` were all `href="#"`, flagged in the 2026-07-25 UI audit and still open. `Documentation` now points at the docs site (**verified 200** before linking); `Terms` and `Privacy` have no destination and are removed — a link to nowhere costs more trust than an absent one, and inventing placeholder pages is out of scope.
- **Scroll anchor moved.** Hero scrolls to `#landing-stats`, which lived in `Talk`. The band takes it so the first scroll lands on the value proposition instead of past it, and `Talk` gives it up — two elements sharing one id is a silent mis-scroll, not an error.

## The bundle rebuild

`dashboard/frontend/index.html` is **not** Vite output. It carries ~370 hand-written lines Vite cannot emit (the auth-gate script, `#landingAuthModal`, `<style id="landing-auth-patch">`, the delegated `[data-landing-auth]` handler). Copying `dist/index.html` over it turns all seven CTAs into dead buttons with **no console error and a page that still renders** (#225).

So only the two asset-reference lines were edited — the `index.html` diff is literally 2 lines. Evidence the rebuild is sound:

- **Reproducible**: a second `npm run build` emits identical filenames *and* identical `sha256` for both bundles, so the committed bundle provably corresponds to the committed source.
- **Integrity guards**: 6/6 pass, including `test_hand_written_auth_layer_survives_a_bundle_refresh` and `test_no_orphaned_assets`. Source and bundle both emit **7** `data-landing-auth` CTAs (was 6; the band adds one — the count is derived at run time, so no assertion was edited).
- **Browser-verified** with Playwright against a backend on a scratch `DATABASE_PATH`: the band renders above Talk, holds the single `#landing-stats` anchor, scrolling to that anchor puts the band in view, 7 CTAs are live in the DOM, and clicking the band's CTA opens the signup modal (`display: none` → `flex`). The only non-2xx request is `/_vercel/insights/script.js`, expected off Vercel.

Two guard docstrings that still said "six" CTAs were corrected — comment-only; the assertions were already dynamic.

## Forbidden claims are pinned as tests

`test_band_makes_no_paper_trading_claim` and `test_band_makes_no_real_capital_claim` fail the build if the copy ever promises either. Both would contradict the code: **no order-submission route exists** — `AlpacaPaperTradingClient` has no order method, `PaperTradingSession.add_trade` has zero production callers, `execution/paper_backend.py` raises `NotImplementedError`, and `ROBINHOOD_EXECUTE` defaults to `false`.

## Test plan

- [x] `pytest dashboard/backend/tests/test_landing_value_band.py -q` — 15 passed
- [x] `pytest dashboard/backend/tests/test_frontend_bundle_integrity.py -v` — 6 passed
- [x] `pytest dashboard/backend/tests/ -q` — 2050 passed, 1 failed, 64 skipped. The single failure (`test_ifind_engine_resolves_csi300_sample20_and_records_provenance`) is **pre-existing and unrelated**: `main` produces the identical failure at 2035 passed (delta is exactly this PR's 15 new tests), and it passes in isolation on both refs.
- [x] Playwright CTA/modal check, 8/8 substantive checks

## Self-review — one CodeQL fix beyond the plan

The plan's `test_footer_documentation_points_at_the_docs_site` asserted the bare host substring, which CodeQL flags **high** as `py/incomplete-url-substring-sanitization` (alert 1185). Fixed rather than dismissed — this repo has already seen a dismissal spawn a fresh alert the moment line numbers shifted — and the stricter form is worth having regardless: the bare-host assertion accepted `...readthedocs.io.attacker.example`, which the pinned full href rejects (mutation-tested).

Worth noting for future scans: that alert appeared **only on `refs/pull/276/merge`**. `refs/heads/feat/landing-value-band` reported zero open alerts the whole time, so a branch-ref query alone would have reported a clean bill of health. Post-fix, the merge ref is back to 0 open with the repo-wide count unchanged at 41.

Phase C of `docs/superpowers/plans/2026-08-01-user-feedback-ux-round.md`. Independent of Phase A (#275) — no shared files.


FlyM1ss:
Review findings addressed in `2c554a8` + `076a056`.

**Blocking — scroll anchor.** `scroll-margin` is read off the element `scrollIntoView()` targets and is not inherited, so the section's `scroll-mt-40` did nothing for Hero's gesture at the zero-height `#landing-stats` child. Measured in Chromium against the shipped bundle:

| | before | after |
|---|---|---|
| 1440x900 | 80 of 80 headline px hidden | 0 |
| 390x844 | 123 of 144 hidden | 0 |

`main` only looked clean because Talk's `items-center` grid offset its heading 126px past the 122.5px chrome — the bug was latent there, and moving the anchor onto a flush-topped section exposed it. All four nav links re-measured at 0px hidden.

**Also fixed**
- `test_hero_scroll_anchor_clears_the_fixed_chrome` — the old id-count guard proved uniqueness, not that the scroll lands anywhere readable. Cross-checks the anchor's `scroll-mt-*` against `--landing-chrome-height`; strictly greater, since that var is a rounded proxy for a 122.5px chrome.
- Forbidden-claim guards now match patterns — the literal list let the natural singular `paper trade` through.
- `#why` linked from the navbar; an id nothing references is decorative.
- Dropped the band's `01/02/03` so the page runs one numbered sequence instead of two.

**Deferred:** removing the dead `Privacy`/`Terms` stubs leaves no privacy/terms surface anywhere while 7 CTAs collect email + password — filed as #278 rather than grown into this PR.

18 landing guards + 6 bundle-integrity pass; full backend suite 2053 passed. CSS hash is unchanged (`scroll-mt-40` was already emitted for Talk/Test/Race); `index.html` took a one-line asset-ref edit with the hand-written auth layer untouched.

PR #281, opened 2026-08-01
fix(auth): rate-limit login and signup attempts

Allan-Feng:
## Summary
- Best-effort in-process rate limits on `POST /api/auth/login` and `/signup` (per client + per email), with `Retry-After`.
- Login failures return `Invalid email or password.` **and** cost the same bcrypt compare whether or not the address exists (was 0.06 ms vs 182 ms — the timing answered what the copy did not).
- Both login budgets meter **failed** attempts only, so a correct password is never refused because of someone else's guesses.
- Per-client keys read the left-most `X-Forwarded-For`: uvicorn rewrites `request.client` only when the peer is in `--forwarded-allow-ips` (default `127.0.0.1`), which Render's router is not, so otherwise every landing-page visitor shares one bucket.
- Env knobs in `.env.example`, each echoed at startup; `0` disables a budget. Agent API-key routes untouched.

## Known gap
`POST /api/auth/signup` still answers "does this address have an account?" — 409, one request. The `/login` change does not cover it and the new budgets do not bound it (one request is enough, and the keys are caller-supplied). Closing it means confirm-by-email signup, which is a product change.

## Not a security boundary
Per-client budgets are keyed on headers the caller controls, so they bound naive abuse and cap queued bcrypt work — not an attacker. What bounds password guessing is the per-email failure counter, keyed on the account under attack. See the `rate_limit` module docstring.

## Test plan
- [x] `pytest dashboard/backend/tests/ -q` — 2093 passed, 64 skipped (1 pre-existing unrelated failure, `test_ifind_engine_resolves_csi300_sample20`, red on this branch's base too)
- [x] Timing re-measured after the fix: 185.3 ms known vs 184.3 ms unknown = 1.01x (was 3112x)
- [x] CodeQL on `refs/pull/281/merge`: 0 open alerts
- [ ] Manual: hammer login from one IP → 429; correct password still works after the email failure budget is spent

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -197,13 +197,21 @@ ACCOUNT_EMAIL_FROM=noreply@example.com
 ACCOUNT_EMAIL_FROM_NAME=Agentic Trading Lab
 
 # Auth rate limits (optional). Best-effort in-process budgets for
-# POST /api/auth/login and /signup. Defaults shown. Per-email login counters
-# meter *failed* attempts only so an attacker cannot lock out a correct password.
-# AUTH_LOGIN_IP_MAX=30
+# POST /api/auth/login and /signup; defaults shown, and each is echoed at
+# startup as "auth rate limit AUTH_...: N per Ns" so the effective value is
+# readable in the deploy log. Set any *_MAX to 0 to disable that budget.
+#
+# Both login budgets meter *failed* attempts only, so a correct password is
+# never refused because of someone else's guesses. The per-IP budgets are loose
+# on purpose: they are keyed on caller-supplied headers (X-Browser-Id, then
+# X-Forwarded-For), so they bound naive abuse rather than a real attacker, and
+# a tight value mostly punishes shared addresses. AUTH_LOGIN_EMAIL_MAX is the
+# one that actually bounds password guessing -- keyed on the targeted account.
+# AUTH_LOGIN_IP_MAX=60
 # AUTH_LOGIN_IP_WINDOW_SECONDS=900
 # AUTH_LOGIN_EMAIL_MAX=10
 # AUTH_LOGIN_EMAIL_WINDOW_SECONDS=900
-# AUTH_SIGNUP_IP_MAX=10
+# AUTH_SIGNUP_IP_MAX=60
 # AUTH_SIGNUP_IP_WINDOW_SECONDS=3600
 # AUTH_SIGNUP_EMAIL_MAX=5
 # AUTH_SIGNUP_EMAIL_WINDOW_SECONDS=3600
--- a/.env.example
+++ b/.env.example
@@ -196,6 +196,18 @@ BREVO_API_KEY=your_brevo_api_key_here
 ACCOUNT_EMAIL_FROM=noreply@example.com
 ACCOUNT_EMAIL_FROM_NAME=Agentic Trading Lab
 
+# Auth rate limits (optional). Best-effort in-process budgets for
+# POST /api/auth/login and /signup. Defaults shown. Per-email login counters
+# meter *failed* attempts only so an attacker cannot lock out a correct password.
+# AUTH_LOGIN_IP_MAX=30
+# AUTH_LOGIN_IP_WINDOW_SECONDS=900
+# AUTH_LOGIN_EMAIL_MAX=10
+# AUTH_LOGIN_EMAIL_WINDOW_SECONDS=900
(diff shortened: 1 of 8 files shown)


FlyM1ss:
Reviewed and pushed the fixes to this branch (`601b4ab`, `f26e0db`). Body updated to match. Diff is the thing to read; short version of what changed and why:

**Blockers**
1. **Per-IP budget was a cap on sign-ins, not on guesses.** `allow()` ran before `authenticate()` and charged successes. Split into `check()`/`record()` and meter failures only, keeping the pre-bcrypt gate. The key was also collapsing: uvicorn rewrites `request.client` from `X-Forwarded-For` only when the peer is in `--forwarded-allow-ips` (default `127.0.0.1`), which Render's router is not — so on prod every landing-page visitor shared one bucket and the effective limit was 30 logins / 10 signups *for the whole site*. `client_ip()` now reads the header directly. Defaults 30→60 and 10→60.
2. **Generic 401 copy didn't close enumeration.** `authenticate()` returned before bcrypt on an unknown email: 0.06 ms vs 182 ms measured, so the response *time* still answered it. Both store twins now go through `verify_password_for_account()`; re-measured at 1.01x. That makes a miss cost ~190 ms of CPU, so `authenticate()` (and signup's `create_user`) moved onto a worker thread — otherwise it's a single-threaded stall anyone can drive, and the limiters can't bound it when rotating a header buys a new budget.
3. **`/signup` still leaks account existence** via 409 in one request. Left as-is, documented at the raise site and in the body rather than implied closed.

**Also:** `print()` over `logger.info` (invisible under the deployed uvicorn config, and `extra={}` is never rendered anyway); `_env_int` takes `0` as "disabled" and reports bad overrides, with each effective limit echoed at startup; password policy validated *before* the signup budget is charged, so retrying a weak password no longer burns it; `retry_after_seconds` prunes before reading.

**One to flag:** CodeQL had 3 `py/log-injection` alerts on the `logger.info` calls and marked them **fixed** when they became `print()` — it doesn't model `print` as a log sink. The taint was unchanged, and rendering the field for real is what armed it: `_normalize_email` only strips the *ends*, so `victim@example.com\nauth.login_failed domain=attacker.test` validates and writes a forged entry into the log. Fixed properly in `f26e0db` (truncate at the first non-domain character) with a test.

15 new tests. Full suite 2093 passed; CodeQL 0 open on `refs/pull/281/merge`. The one red test (`test_ifind_engine_resolves_csi300_sample20`) is red on this branch's base too — unrelated.

FlyM1ss:
Merged as `e5ecc55`, deployed, and verified live: 10 × 401 on a nonexistent address then 429 on the 11th, `Retry-After` counting down 874 → 872 inside the 900s window. That is the per-email budget (10) tripping well before the per-IP one (60), which is the intended shape.

Two notes for whoever touches this next.

**`FORWARDED_ALLOW_IPS=*` is now set on the Render web service.** Not required by this PR — `client_ip()` reads the header itself, and both paths take the same left-most `X-Forwarded-For` entry, so they agree rather than diverge. I set it because the untrusted-peer path skips more than `scope["client"]`: it also skips `scope["scheme"]`, which had been leaving prod on `http`. That fed `request.base_url` in two places that build user-facing URLs — `strategies.py::_share_base` (share links; it reads `PUBLIC_BASE_URL`, which is unset in prod — the var that *is* set is `PUBLIC_APP_URL`, a different name) and the `client_hint` SDK snippet in `agents.py`. Both were emitting `http://`.

Verified against uvicorn 0.44.0 with `X-Forwarded-For: 203.0.113.7, 10.11.12.13`:

| `FORWARDED_ALLOW_IPS` | `client.host` | `scheme` |
|---|---|---|
| peer not in list (prod before) | the peer | `http` |
| `*` (prod now) | `203.0.113.7` (left-most) | `https` |
| the peer's own IP | `10.11.12.13` (right-most untrusted) | `https` |

Worth knowing that `*` short-circuits to the **left-most** entry, which is caller-supplied — it buys granularity, not trust. That is the same tradeoff `client_ip()` already documents, and the unspoofable alternative (narrow trust, right-most untrusted) needs a stable router IP that Render doesn't publish.

**`client_ip()`'s docstring is now stale** where it says Render's router is not in `--forwarded-allow-ips`. The reasoning for doing it app-side still stands; only that one factual claim aged out the moment the var was set.

PR #283, opened 2026-08-01
fix(deploy): same-origin API via Vercel rewrites

Allan-Feng:
## Summary
- Frontend `API_BASE` is now `''` off localhost (root-relative), instead of hardcoded `https://agentictrading.onrender.com`.
- `vercel.json` rewrites `/api`, `/paper`, `/backtest`, `/runs`, `/config`, `/admin`, `/ticker`, `/health`, `/compare` to Render.
- Optional `ATL_FRONTEND_ORIGINS` CORS allowlist (still `*` when unset; credentials remain off).
- Agent API-key clients can keep calling Render directly; this is for the browser app.

Prerequisite for HttpOnly session cookies (follow-up). Landing Vite bundle still embeds its own fetch base only in the auth modal via `index.html` (updated).

## Test plan
- [x] `pytest dashboard/backend/tests/test_frontend_api_base.py dashboard/backend/tests/integrations/test_strategy_html_api_base.py -q`
- [ ] After Vercel deploy: DevTools Network shows `/api/...` same-origin; responses come from rewrite
- [ ] Local `uvicorn` on :8000 still works for `/` and `/app`

PR #284, opened 2026-08-01
fix(auth): move browser sessions to HttpOnly cookies

Allan-Feng:
**Gate cleared 2026-08-04:** #301 resolved (#302 + #303) — `curl -sI https://agentic-trading-lab.vercel.app/api/health` now returns 200 from Render through the Vercel rewrite. Merge-from-main redone and review fixes applied (see comments); suite green at 2320 passed.

---

## Summary
- Set an HttpOnly `atl_session` / `__Host-atl_session` cookie on login/signup; stop returning `token` in JSON and stop writing `auth-token` to `localStorage`.
- Frontend fetches use `credentials: 'include'`; CORS enables credentials only when `ATL_FRONTEND_ORIGINS` is set (never with `*`).
- Bearer remains a deprecated bridge for scripts/tests. Depends on hashed sessions (#282) and same-origin Vercel proxy (#283).

## Test plan
- [ ] `pytest dashboard/backend/tests/test_auth.py` — cookie-only `/me`, no `token` in signup/login JSON
- [ ] Local: sign up / log in on `/` and `/app`; DevTools Application → Cookies shows HttpOnly session; localStorage has no `auth-token`
- [ ] After #283: production Vercel origin can call `/api/auth/me` with credentials (or same-origin rewrite)
- [ ] Agent `X-API-Key` flows unchanged (no cookie/CSRF yet on those routes)


FlyM1ss:
Blocked by #301 (Vercel project does not apply the repo-root `vercel.json`). Converted to draft; see the first line of the body.

FlyM1ss:
## Review — sound migration, but four things must land with it

Reviewed against `main`. Suite green at 2078 passed / 64 skipped. #284's own delta is `1294756..HEAD` (20 files, +309/−120); everything else in `gh pr diff 284` belongs to #282 (already squash-merged) or #283.

### Blockers

**1. Merge hazard — `main` is strictly ahead of this branch's copy of #282.**
This branch carries #282's *original* commit `a743fd9`; `main` has the squash (`f463052`) plus its review fixes. A resolution that keeps "the branch side" silently reverts merged, deployed hardening. Missing from the branch and present on `main`: `require_session_hash_secret()` (the startup gate that fails a prod boot with no `SESSION_HASH_SECRET` instead of 500-ing every later request), `_report_once` print-based env announcements (the branch uses `logger.warning`, which emits **nothing** under deployed uvicorn), `_best_effort_write`, `verify_password_for_account` + `_dummy_password_hash` (constant-time login — closes a ~3000× timing user-enumeration oracle), `purge_expired_sessions`, `idx_auth_sessions_expires_at`, and the single `AUTH_SESSIONS_DDL` literal the twin-parity guard reads by source text.
*Resolution:* take `origin/main` wholesale for `session_tokens.py`, `users.py`, `users_postgres.py` — #284's own commits don't touch them. **`tests/test_users_postgres.py` is the exception** — #284 *does* touch it, so take `main`'s version and re-apply this PR's delta (purely the "token no longer in JSON → read from the cookie jar" adaptation). Also union, don't pick, in `app.py` (GZip must stay the **first** `add_middleware` call = innermost) and `test_app_composition.py` (`main` added the rejected-orders route to two golden sets and rewrote `test_middleware_order_preserved`).

**2. `app.html:1645-1649` — cache-busters not bumped, so signed-in users render signed-out.**
Four frontend files changed, none bumped. `home-page.js`, `js/portfolio.js` and `js/agent-editor.js` fall under `vercel.json`'s catch-all `max-age=3600`. So for up to an hour a browser runs the **new** `app.js` — which stops writing `localStorage['auth-token']` — beside the **old** helpers, all three of which still gate on that key: `home-page.js:752` `isHomeSignedIn()` → Home paints the signed-out CTA; `js/portfolio.js:52` → My Portfolio shows the sample-data badge instead of the real ledger; `js/agent-editor.js:314` → "Sign in required", Connect Robinhood disabled. All while genuinely signed in. Note the existing guards can't catch this — `test_editor_asset_cache_bust_advances_with_its_source` and `test_cache_bust_versions_were_bumped` assert `>= floor`, not "changed", by design.

**3. Every currently-signed-in user is logged out on deploy.**
The new `app.js` never sends `Authorization: Bearer` again (`API.request`, `AuthAPI.request` now pass only `credentials: 'include'`) and no cookie exists yet, so boot `refreshAuthUser()` → `GET /api/auth/me` → 401 → `clearAuthState()`. Their server-side session row stays alive and unreachable until it expires. A ~6-line bridge avoids it: read the legacy `localStorage['auth-token']` once in `refreshAuthUser()` and send it as Bearer on that single `/me` call; in `me()`, `set_session_cookie(response, token)` when the token arrived via the header. Strictly additive and fails safe — if it doesn't work the user gets the same re-login they'd have gotten. A forced re-login may be an acceptable price, but it should be a decision, and the PR body doesn't mention it.

**4. `vercel.json` catch-all caches authenticated JSON.** `"source": "/(.*)"` → `Cache-Control: public, max-age=3600` also matches `/api/:path*`, and the backend sets no `Cache-Control` anywhere, so the config value is the only one on the response. This PR is what turns those responses into credential-bearing ones: sign out on a shared laptop, next visitor's boot `GET /api/auth/me` is served from disk cache, UI paints signed-in as the previous user. *(Attribution: the rule comes from #283, not this PR's commits — and #283 is now reverted, so re-check once it re-lands.)*

### Major

- **`auth_cookies.py:56-60` — `clear_session_cookie` cannot delete the production cookie.** `Response.delete_cookie` defaults `secure=False` (verified against installed starlette 0.47.2), and RFC 6265bis §4.1.3.1 makes a UA **reject** any `Set-Cookie` whose name starts with `__Host-` and lacks `Secure`. Measured with `ATL_COOKIE_SECURE=true`: the emitted header is `__Host-atl_session=""; …; Path=/; SameSite=lax` — no `Secure`, no `HttpOnly`. Not a bypass (`delete_session` revokes server-side), but the function's comment claims a guarantee it doesn't deliver, and it fails *only* in production. Pass `secure=cookie_secure(), httponly=True, samesite="lax"`.
- **The cookie path is barely tested.** `test_me_accepts_session_cookie_without_bearer` is the only case where a cookie actually authenticates anything; ~8 modules were mechanically rewritten to pull the raw token *out of the cookie jar* and keep sending it as Bearer (`test_discord_linking.py:45-48` even re-synthesises the `token` key the API no longer returns). Uncovered: `_optional_user`'s new cookie path, logout clearing the cookie (zero assertions), and the whole `cookie_secure()=True` branch — which is exactly why the bug above survived. Assert on the raw header, not the jar: TestClient over `http://testserver` drops `Secure` cookies.

### Minor

`app.py:72-82,91` — `_cors_allow_origins()` unconditionally appends the four loopback origins, and this PR pairs that list with `allow_credentials=True`; append them only when `cookie_secure()` is false · `api/auth.py:243` — `POST /api/auth/logout-all` has no caller, no UI and no test · `conftest.py:50` — pins `SESSION_HASH_SECRET` but doesn't strip `ATL_COOKIE_SECURE`/`ATL_ENV`/`RENDER`, which now decide the cookie name and `Secure` flag; with any set in the shell, 8 modules fail with a misleading "missing session cookie" · `test_auth.py:16-21` duplicates `auth_cookies_helpers._cookie_session_token` verbatim.

### Checked and refuted — please don't re-raise

A `Domain`-scoped cookie being dropped through the Vercel→Render proxy (`set_session_cookie` passes no `domain`; measured header is host-only and satisfies every `__Host-` requirement on the set path) · landing page blocked by `Access-Control-Allow-Origin: *` under `credentials: 'include'` (both inline scripts compute `API_BASE = ''`, so it's same-origin) · stale cached JS not sending the cookie (`fetch` defaults to `credentials: 'same-origin'`) · shipping cookie auth before #285 opening a CSRF hole (`SameSite=Lax` withholds the cookie cross-site on non-safe methods; every session-authenticated `GET` enumerated and all are read-only) · SDK/Discord bot breakage (both `X-API-Key`) · Postgres twin parity (this PR's own commits touch neither file) · logout leaving the session usable (server-side revocation is correct; only the cookie deletion is broken).

---

Blocked on #301 — #283 was reverted because the Vercel rewrites don't apply on the live project, and without same-origin this change is strictly worse than the status quo.

FlyM1ss:
Platform blocker cleared: #301 is resolved (#302 + #303 — same-origin API rewrites are verified live in prod). Before this can merge it still needs its merge-from-main redone against current `main` (now includes #294–#296 and #303) per the per-file rules in the review comments, plus the `clear_session_cookie` secure-flag fix.

FlyM1ss:
Pushed two commits to get this merge-ready now that #301/#303 are live:

**`ef38d8b` — merge from main** (main moved a lot since this branched: #282 squash + review fixes, #292–#296, #302, #303). Per-file resolutions follow the review's rules: `session_tokens.py` / `users.py` / `users_postgres.py` taken wholesale from main (keeps `require_session_hash_secret`, constant-time login, `purge_expired_sessions`, the twin-parity DDL literal); `api/auth.py` unions main's rate limits/threadpool/telemetry with the cookie issuance; `app.py` keeps GZip as the innermost middleware and gates `allow_credentials` on `ATL_FRONTEND_ORIGINS`; root `vercel.json` deleted (rewrites live in `dashboard/frontend/vercel.json` since #302).

**`7a3344f` — review fixes:**
- `clear_session_cookie` now sends Secure/HttpOnly/SameSite on the `__Host-` deletion — without it the prod cookie survived logout in the browser (RFC 6265bis rejects the unflagged delete).
- Migration bridge (review blocker 3): app.js sends the legacy localStorage token once as Bearer on the boot `/me`; the backend answers with Set-Cookie, so everyone signed in today stays signed in through the deploy instead of being force-logged-out.
- Cache busters bumped for all four changed frontend files (review blocker 2 — stale helpers rendered signed-in users as signed-out for up to an hour).
- Raw Set-Cookie header tests for the Secure branch (TestClient's jar silently drops Secure cookies, which is exactly why the logout bug survived), plus the bridge test; conftest strips `ATL_COOKIE_SECURE`/`ATL_ENV`/`RENDER`.
- Three main-side tests adapted to read the token from the cookie jar (it left the JSON body).

Review blocker 4 (the vercel.json catch-all caching `/api` responses) is already closed by #303's `no-store` rule on all proxied paths.

Suite: 2320 passed / 66 skipped. Marked ready for review.

PR #285, opened 2026-08-01
fix(auth): add CSRF checks for cookie sessions

Allan-Feng:
~~DO NOT MERGE~~ **Unblocked 2026-08-04**: #283 re-landed via #303, #284 merged, `main` merged into this branch (`eb6edd9`) with the pre-merge fixes: login/signup exempt from the double-submit gate (stale-cookie lockout), CSRF tokens compared as UTF-8 bytes (non-ASCII header 500), and `ATL_FRONTEND_ORIGINS` set in Render. CI + CodeQL green, 0 open code-scanning alerts.

This PR sits on top of #284's branch, so merging it alone would drag the whole HttpOnly-cookie change in with it. The CSRF guard is only meaningful once cookie sessions exist.

Re-open for review once `curl -sI https://agentic-trading-lab.vercel.app/api/health` returns 200 from Render rather than `x-vercel-error: NOT_FOUND`.

---

## Summary
- Double-submit CSRF cookie (`atl_csrf` / `__Host-atl_csrf`) + `X-CSRF-Token` on mutating requests that carry the account session cookie.
- Reject disallowed `Origin`/`Referer` against `ATL_FRONTEND_ORIGINS` (+ local hosts / request host).
- Frontend attaches CSRF headers on credentialed fetches. Depends on HttpOnly cookies (#284).

## Test plan
- [ ] `pytest dashboard/backend/tests/test_csrf.py`
- [ ] Confirm agent `X-API-Key` mutating calls without a browser session cookie still work
- [ ] Browser: after login, logout/change-password succeed; request without `X-CSRF-Token` returns 403


FlyM1ss:
Blocked by #301 (Vercel project does not apply the repo-root `vercel.json`). Converted to draft; see the first line of the body.

FlyM1ss:
## Review — CSRF design is sound, but two blockers lock users out

Reviewed **#285's own delta only** (`git diff allan-fork/feat/auth-httponly-cookie HEAD` — 9 files, 329 insertions); #282/#283/#284's content is reviewed separately. Suite green at 2082 passed / 64 skipped. Findings below were each verified by driving the app, not by reading alone.

**Design classification:** naive double-submit, not a synchroniser token — a floating 256-bit `secrets.token_urlsafe(32)` with no server-side record and no session binding. Defensible here (`__Host-` blocks Domain-scoped writes, both `vercel.app` and `onrender.com` are on the Public Suffix List, session cookie is `SameSite=Lax`, Origin allowlist is a second layer). But `session_tokens.py` already owns an HMAC key, so `HMAC(session_hash_secret, session_digest)` is a one-line upgrade that closes cookie-injection as a *class*. Worth doing while it's cheap.

### Blockers

**1. `csrf.py:110-117` — `ATL_FRONTEND_ORIGINS` is unset in prod, so every mutating request 403s.**
`allowed_browser_origins()` would contain only the four localhost origins plus `request.base_url` (the Render host, not the Vercel host). Verified with enforcement on and the var unset: `POST /api/auth/login` with `Origin: https://agentic-trading-lab.vercel.app` → `403 {"detail":"Cross-origin request blocked"}`. That is login, signup, agent create, backtest start — the whole product.
*Fix:* set `ATL_FRONTEND_ORIGINS=https://agentic-trading-lab.vercel.app` in the Render dashboard **before** merging (single-key PUT; env writes don't redeploy, so it's free to do first). Note the same var also drops CORS off `*` and flips `allow_credentials=True` — one switch controlling two subsystems. Vercel preview domains (`*-git-*.vercel.app`) will still 403 unless added.

**2. `csrf.py:164` + `index.html:145-150,291-297` — a user returning after a day cannot log back in.**
The gate keys on session-cookie *presence*, not validity. Sessions idle out at **24h** while the cookie lives **7 days** (`session_tokens.py:20-21`), so the stale-cookie state is the common return-visit case. `/api/auth/me` 401s without clearing the cookie, the cookie is HttpOnly so JS can't clear it, and the landing's `clearAuth()` only clears localStorage. The user then POSTs `/api/auth/login` with a stale cookie and no token → verified `403 {"detail":"CSRF token missing or invalid"}` (the same request with no cookie gives a normal `401 Invalid email or password`). The landing only fires its bootstrap `/api/auth/me` when `localStorage[auth-user]` exists, so the middleware's back-fill never runs either. Dead end with no user-recoverable path.
*Fix:* exempt `/api/auth/login` and `/api/auth/signup` from double-submit — they *establish* a session, the only threat is login-CSRF, and the Origin check already covers it.

### Major

- **`csrf.py:133` — unauthenticated 500.** `secrets.compare_digest` raises `TypeError` on non-ASCII `str`; Starlette decodes headers as latin-1, so a raw `0xE9` in `X-CSRF-Token` becomes `'é'`. Verified via the ASGI app directly → unhandled `TypeError` → bare 500, no account needed. Compare bytes instead. (With `CORSMiddleware` innermost that 500 escapes un-headered — the "prod CORS errors mask bare 500s" shape.)
- **`strategy.html:156-162` — logged-in users can't run a strategy backtest.** The page has one inline script, never loads `app.js`, so `window.csrfHeaders` doesn't exist; `POST ${API}/backtest/run` is same-origin so the cookie attaches but no token does. Verified → `403`. Needs its own `readCsrfToken()`.
- **`csrf.py:155-159` — the Origin gate runs before the API-key skip.** Verified: `POST /api/v1/agents` with `X-API-Key` + `Origin: https://evil.example` → `403 Cross-origin request blocked`. Browser-based API-key integrations work today (CORS `*`, `x-api-key` allowed) and would break. The stdlib SDK and Discord bot send no `Origin` and are unaffected — both checked. Decide deliberately: move the API-key short-circuit above `origin_allowed`, or document the behaviour in `agent_api.rst`.
- **`conftest.py:52-56` — `ATL_CSRF=0` process-wide makes 2082 of 2086 tests run with the middleware inert.** CI structurally cannot catch a CSRF regression; that is exactly why both blockers and the `strategy.html` break are invisible to a green suite.
- **`test_csrf.py:71-83` — `test_x_api_key_alone_skips_csrf` proves nothing.** The request 400s on missing owner context so the key is never validated, *and* the assertion `status != 403 or "CSRF" not in detail` also passes when the middleware genuinely blocks (`"Cross-origin request blocked"` contains no `"CSRF"`). Both verified.

### Minor

`csrf.py:49` hardcodes `max_age` instead of deriving from `session_ttl_days()` · `csrf.py:151` back-fills a CSRF cookie for any `atl_session` value, including on 401s · `csrf.py:136-139` `ATL_CSRF` is a silent fail-**open** kill switch — `print()` the mode like every store factory does (`logger.info` is invisible under deployed uvicorn) · `csrf.py:125` duplicate case-insensitive header lookup is dead code · `agent-editor.js:820-831` the `window.API?.patch` fallback still builds headers without the token · `app.html:1646-1647` cache-busters not bumped (watch for collision with #295/#296) · `CsrfMiddleware` sits in `csrf.py` rather than `middleware.py` where its siblings live.

### Checked and refuted — please don't re-raise

Planted unprefixed `atl_csrf` overriding `__Host-` (checked first, verified 403) · OAuth GET callbacks as a linking hole (both validate HMAC-signed `state`; Robinhood writes nothing) · guard only covers `/api/auth` (it's app-level; verified 403 on `/backtest/run`, `/api/v1/agents`, `DELETE /api/v1/agents/{id}`, `/api/v1/portfolio/allocate`) · SDK/Discord bot breakage (neither sets `Origin` or cookies) · `_request_self_origin` as a spoofable bypass (browser sets `Host`; uvicorn honours `X-Forwarded-Proto`/`-For`, not `-Host`) · Bearer callers wrongly gated (gate keys on the cookie only) · missed route-contract golden set · localhost origins exploitable (`Origin` is browser-controlled) · token entropy/comparison (256-bit, constant-time, rotated on signup/login, cleared on both logouts).

---

Separately: this is blocked on #301 regardless — #283 was reverted because the Vercel rewrites don't apply on the live project.

FlyM1ss:
Platform blocker cleared: #301 is resolved (#302 + #303). Still gated on: #284 landing first, the login-lockout fix (exempt login/signup from the CSRF gate), the non-ASCII `compare_digest` 500, and `ATL_FRONTEND_ORIGINS` being set in the Render dashboard before merge.

PR #288, opened 2026-08-02
fix(backtest): bound T+1 rejection audit

FlyM1ss:
Review follow-ups on #272. Targets `feat/ifind-ashare-t1`, so merging this updates #272 rather than `main`.

**Blocking finding.** `rejected_orders` is unbounded per-step audit data, and `RunMetadata` is the `response_model` for two *list* routes — `/api/backtest/runs` (fetched 3× by `app.js` on load) and the public unpaginated `/runs`. ~198 B/record × a plausible 7,200-record A-share run ≈ 1.4 MB in one `agent_runs.metadata` cell and ~14 MB per list response, into free-tier `ATL-runs-main`. Now: capped head sample + `rejected_orders_count` + `rejected_orders_truncated`; records served by a new `GET /runs/{run_id}/rejected-orders`; `RunMetadata` keeps only the two scalars. Live progress carries a tail window instead of rewriting the whole list every step.

**The rule was unobservable.** `get_portfolio_state` exposed only total positions, so both decision paths proposed the full holding and every bar of a buy date minted a `t1_frozen` record for a constraint the agent was never shown. Adds `sellable_shares` (T+1 only — non-A-share prompts are byte-identical) and sizes sells against it.

**Settlement now resolves from the profile registry** at all three `PortfolioManager` sites, not the constructor default, with an AST guard so a fourth cannot omit it. `external_run_service` is Alpaca-only today; defaulting there would have made a future iFinD run settle same-day with no error and no log.

Also: tolerance on the unfilled-share test (float residue minted a phantom rejection of ~2.8e-17 shares), the two non-T+1 divergences of the rewritten SELL branch documented, the `py/unnecessary-lambda` CodeQL alert on #272's merge ref cleared, spec renamed to `.zh-CN.md`.

2066 passed / 64 skipped (2045 at #272 head; +21 new tests). Ledger invariant re-verified over 35,886 randomized actions, 0 violations. The epsilon guard and the AST guard were both mutation-checked.

## Second commit (`e21751f`) — self-review findings

Reviewing my own first commit turned up two real defects and I fixed both.

- **`sellable_shares` was write-only.** The LLM prompt builds each holding from a five-key whitelist and dropped it, and no prompt channel mentioned T+1 at all — so the model could not see the rule its orders were capped against. Now forwarded into `current_holdings`, with `settlement`/`settlement_note` in `market_context` naming the rule. Mutation-checked: removing the forward fails the new test.
- **Capping traded a too-loud audit for a silent one.** A capped sell fills exactly, so the executor records nothing, and a fully-frozen holding emits no action at all — *"how often did T+1 stop this agent exiting?"* became unanswerable. Capped sells now report the discarded intent: a `t1_deferrals` ledger deduped per `(symbol, trading day)` (bounded by symbols × days, not per bar), `t1_deferred_events`/`t1_deferred_shares` on the run, and `requested_shares` on the action so the decision log doesn't read as though the agent only ever wanted the capped size.
- Plus: the AST guard's three blind spots documented, `sellable_positions` returned as a read-only view, head-vs-tail sample rationale recorded.

---

⚠️ **The checks list on this PR is empty because it targets a feature branch, not `main`.** Both `ci.yml` and `codeql.yml` are `pull_request: branches: [main]`, so neither fires here — empty checks mean *never ran*, not *passing*. Dispatched manually against the branch instead (`deploy-prod` is gated on `github.ref == 'refs/heads/main'`, so nothing could deploy):

| Run @ `e21751f` | Result |
|---|---|
| CI → Backend tests | ✅ **2126 passed, 7 skipped** (skip collapse from 64 local confirms the `@pg_only` tier ran) |
| CI → Packaging tests | ✅ 171 passed, 2 skipped |
| CI → Trigger Render prod deploy | skipped (correct — not `main`) |
| CodeQL python + js/ts | ✅ scanned `e21751f`; **0 alerts introduced** |

**CodeQL / GHAS status:** the PR-scoped (merge-ref) view is **0**. The branch carries 11 open alerts, but none of them lives in any file this PR touches — they are `main`'s inherited backlog and are exactly the set PR #279 already fixes (its merge ref is 0). Duplicating them here would collide with #279 at merge. Dependabot 0; secret scanning returns 404 on this repo.

Merging this into `feat/ifind-ashare-t1` re-runs the real gated CI on #272, since that one does target `main`.


FlyM1ss:
Merged into `feat/ifind-ashare-t1` as `ed4ebfe`, ahead of #272 merging to `main` — so the T+1 work reaches prod with the audit bound and the deferral ledger in place.

No outstanding findings: all three review items and the nits were fixed in-branch. CodeQL on the PR merge ref: 0 alerts.

PR #292, opened 2026-08-02
perf: run blocking-I/O routes in the threadpool + ticker SWR

FlyM1ss:
Backend half of the "ATL takes 30+ seconds to load" fix. Safe to merge on its own — no API contract changes (one additive field: `/paper/account` gains the same `cached` flag its sibling routes already have).

**What was wrong (measured in prod):** a `/ticker` cache miss spends ~6s in sequential yfinance calls *on the event loop* — during it, a concurrent `/health` went 0.9s → 6.3s. Every page's requests queue behind whoever misses the ticker cache. 58 handlers across 12 routers were `async def` while doing sync I/O (yfinance/requests/sqlite/sync psycopg).

**Changes**
- Flip those 58 handlers to plain `def` → Starlette threadpool. Zero `await`s existed in any of them.
- Ticker SWR cache: fresh (<30s) → instant; stale (<15 min) → instant + background refresh; cold → single-flighted inline fetch. Empty fetch results never overwrite good stale data, so a yfinance outage degrades to slightly-old prices instead of a blank ticker.
- Cache `/paper/account` (30s TTL) — the only paper route that had no cache.
- `GZipMiddleware(minimum_size=1024)` — equity curves / agent lists compress ~10x.

**Tests:** 4 new files (threadpool pin + timed concurrency proof, SWR behavior incl. single-flight, account cache, gzip). Two existing tests updated: the middleware-order freeze learns GZip; the paper-account sanitization test clears the cache first. Full suite green locally except the known iFinD `.env`-leakage failure (pre-existing, unrelated).

**Left for follow-up:** `auth.py` change-password / email-change still block the loop on bcrypt — not boot-critical.

### DIFF
--- a/dashboard/backend/api/auth.py
+++ b/dashboard/backend/api/auth.py
@@ -362,7 +362,7 @@ async def me(current_user: dict = Depends(get_current_user)):
 
 
 @router.post("/logout")
-async def logout(authorization: Optional[str] = Header(default=None)):
+def logout(authorization: Optional[str] = Header(default=None)):
     token = _extract_bearer_token(authorization)
     if token:
         users_module.user_store.delete_session(token)
--- a/dashboard/backend/api/routers/admin.py
+++ b/dashboard/backend/api/routers/admin.py
@@ -28,7 +28,7 @@ router = APIRouter()
 
 
 @router.delete("/admin/runs/{run_id}")
-async def admin_delete_run(run_id: str, request: Request):
+def admin_delete_run(run_id: str, request: Request):
     """⚠️ Delete a specific run (must be owned by session)."""
     session_id = request.state.session_id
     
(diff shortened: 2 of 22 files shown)


FlyM1ss:
Merged as `4ee4a38`. Review findings addressed in-branch; two deliberately deferred and filed as issues so they survive this PR closing:

- #297 — `auth.py` still runs bcrypt on the event loop (`login`/`signup`/`change_password`/`request_email_change`). `change_password` has no `await` and is a trivial flip.
- #298 — the anyio 40-thread limiter is now the throughput ceiling; `algo_chat` and `backtest/run` can each hold a thread for a long time.

Not filed: `py/print-during-import` in `dashboard/scripts/backtest_hourly_agent.py` is the repo's only open CodeQL alert (severity `note`) and is untouched by this PR.

PR #325, opened 2026-08-07
feat(leaderboard): daily board refresh cron and status UI

JinBoatus1:
## Summary
- Add `POST /api/v1/leaderboard/daily/refresh` (secret-gated) plus a GitHub Actions nightly workflow to deploy competition models for the rolling daily window.
- Surface `daily_status` / pending notice on the Daily Leaderboard tab, and default Competition nav to the contest board.
- Mark a daily window complete only when every model deploy succeeds, so partial failures are retried on the next cron (`force=false`).

## Test plan
- [x] `python -m pytest dashboard/backend/tests/test_leaderboard_api.py -v`
- [ ] Configure `LEADERBOARD_DAILY_REFRESH_SECRET` on Render + matching GitHub Actions secret before enabling the workflow in prod
- [ ] Manual: open Competition → Daily tab and confirm notice / poll while models are pending
- [ ] Manual: `POST /api/v1/leaderboard/daily/refresh?deploy_models=false` with the secret returns 200


Allan-Feng:
Pushed a follow-up commit onto this branch (`93ab4b9`):

- **CI goldens**: `EXPECTED_LEADERBOARD_ROUTES` + `app.js?v=77` / `styles.css?v=86`
- **Async refresh**: `POST /daily/refresh` now returns **202** and runs deploys in a background thread (so Render/GHA HTTP timeouts cannot abort the job)
- Workflow shortened to a fast enqueue; `--remote` script timeout aligned

Please re-check CI, then set `LEADERBOARD_DAILY_REFRESH_SECRET` on Render + Actions before relying on the nightly cron.

FlyM1ss:
Reviewed and pushed two fix commits to this branch (`5c7d2e5`, `26ee54f`). Checks are green — 2666 passed / 7 skipped, merge ref at 0 CodeQL alerts.

**The one that matters.** `get_leaderboard()` schedules a refresh on every request, and `GET /api/v1/leaderboard?period=daily` is public and unauthenticated. `_auto_deploy_daily_models_enabled()` returned `not os.getenv("RENDER")`, so an anonymous GET could kick off `deploy_model_run` for all 7 competition entries — billable calls — on the Docker image, any self-host or fork, and inside the test suite (`conftest.py` strips `RENDER`). Reproduced before fixing. Auto-deploy is now strict opt-in via `LEADERBOARD_DAILY_AUTO_DEPLOY`; please don't restore the old default.

Also in `5c7d2e5`:
- Dropped `allow_fallback` from the endpoint and the whole call chain — it waives the H6 integrity guard, so it shouldn't be reachable over HTTP behind one shared secret. Still available on the local CLI.
- 401 for both unconfigured and wrong secret (a public endpoint shouldn't say whether it's armed); operator signal goes to the log. Rate-limited 20/hr.
- `compare_digest` on bytes — a non-ASCII header was an unhandled 500.
- Replaced a full `get_runs_by_session` scan *per LLM entry* with one indexed pass.
- `thread.start()` failure no longer strands the in-progress flag; status is snapshotted after scheduling so the UI actually polls.

`26ee54f` routes the flag through a single writer, which clears the `py/unused-global-variable` alert the first commit introduced.

**Ready to merge** — I've set `LEADERBOARD_DAILY_REFRESH_SECRET` in Render and as a repo secret, same value, so the cron isn't inert on arrival. `LEADERBOARD_DAILY_AUTO_DEPLOY` is deliberately left unset. Note the Render env-var API doesn't redeploy, so the var goes live with the deploy this merge triggers.

Two things left alone by design: market holidays can leave the notice at "pending", and the in-progress flag is per-process (fine on single-instance Render, would duplicate deploys on multi-replica). Filed #327 for a pre-existing exception-text leak on the sibling GET handler — out of scope here.

PR #326, opened 2026-08-07
fix(leaderboard): clear curve emphasis when pointer leaves a line

Allan-Feng:
## Summary
- Leaderboard equity chart only emphasizes a curve when the pointer is near it (within the plot + hit radius); empty plot space and the right-side label gutter no longer stick on a single series.
- Tooltip uses the same proximity gate; idle view no longer auto-emphasizes the current leader (row click still selects).

## Test plan
- [ ] Open Competition Leaderboard equity chart
- [ ] Hover directly on a curve → that curve emphasizes, others dim, tooltip shows
- [ ] Move pointer into empty plot space → all curves return to equal/global view, tooltip clears
- [ ] Move pointer into the right endpoint-label gutter / outside the canvas → no single-curve sticky emphasis
- [ ] Click a table row → that team stays emphasized until another selection; hover still temporarily overrides

### DIFF
--- a/dashboard/backend/tests/test_frontend_leaderboard_hover.py
+++ b/dashboard/backend/tests/test_frontend_leaderboard_hover.py
@@ -1,0 +1,245 @@
+"""Guards for the leaderboard equity chart's hover gate (js/leaderboard.js).
+
+The chart emphasizes one curve and dims the rest while the pointer is on a line.
+Getting the *gate* wrong is invisible to every other test in the suite, and two
+of the ways to get it wrong are not obvious from reading the code:
+
+* Measuring to the nearest **data point** rather than to the rendered line looks
+  right on the Contest board (162 samples per curve, ~4px apart) and breaks the
+  Daily board (8 samples, ~97px apart), where most of every segment then sits
+  further than the hit radius from either endpoint. `resolveHoverTarget` is
+  therefore exercised below at Daily-board spacing.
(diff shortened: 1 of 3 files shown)


FlyM1ss:
Pushed `eee2065` addressing review findings on this branch. Verified in a real browser (Chromium + the live `/api/v1/leaderboard` payloads for both boards), not by reading alone.

**Gutter was still sticky.** Chart.js only fires `onHover` inside `chartArea`, so the 120px endpoint-label gutter never reached the proximity gate — and `chart.update()` replays the last in-plot event, which re-armed a hover right after clearing it. Chart.js is now out of event handling (`events: []`); a canvas `pointermove` listener drives hover, marker and tooltip.

**Hit test was density-dependent.** Distance to the nearest *data point* works on the Contest board (162 samples, ~4px apart) and misses ~2/3 of every segment on the **Daily** board (8 samples, ~97px apart) with the cursor on the line. Now measured to the interpolated curve.

**`pointHoverRadius` ignored the gate**, marking the nearest curve in empty plot space. Replaced with a marker plugin keyed off the gate.

Plus: `getEmphasisLabel` unused param, hover state cleared when the dataset array is rebuilt, `clearChartHoverEmphasis` resets state with no live chart.

`test_frontend_leaderboard_hover.py` (10 cases) runs the extracted resolver under node at Daily-board spacing and pins `events: []` / pointer listeners, so restoring a Chart.js default reintroducing this fails loudly. Full backend suite: 2611 passed, 67 skipped.

PR #328, opened 2026-08-09
docs: participatory competition pivot spec + plan

FlyM1ss:
Gate cleared by the repo owner on 2026-08-15 after the reconciliation below. (This line previously read "DO NOT MERGE until the team has signed off"; recording that it was lifted deliberately, not overlooked.)

Docs only. No code, no schema, no routes.

**What this is.** ATL's hook becomes a participatory leaderboard: users write one trading instruction, run it against the same harness the house entries use, and see where they land.

**Reconciled 2026-08-15 (`78368ad`).** This was written on 2026-08-09 and described two boards that did not exist. PR #352 then shipped two boards with different names and a different cadence, and PR #357 rewrote the landing page around them — neither against this spec. The docs now match prod:

| Was | Now | Note |
|---|---|---|
| Replay | **Competition Leaderboard** | same window; prod has no entry path |
| Forward Season | **Live Trading Leaderboard** | cadence **weekly → two weeks**; prod has no advance engine |

The design is no longer "add two boards" but **"add user entry to the two boards that now exist"** — strictly less work.

**Three things a reviewer should look at:**

1. *The Phase 0 gate got a better prior.* The plan justified the probe by claiming the prompt is dominated by an unconditional `SAFE_TRADING_PROMPT`. It isn't — a custom instruction **replaces** that body outright (`validator.py:667-711`, joined by `create_custom_prompt` at `:730`). The gate still matters, but a flat result would now be more damning, and the nonsense control is the load-bearing measurement, not the spread.
2. *C8 changed priority.* The Live Trading board is in prod as a Season 0 preview that has never advanced. C8 needs no user entries and doesn't depend on the probe, so it is proposed for splitting out ahead of Phase 2.
3. *One task is now actively wrong.* Task 7 would delete `test_illustrative_example_label_appears_at_least_twice`, a guard #357 added deliberately. Marked superseded rather than silently dropped.

**Also corrected:** `#homeGetStartedBtn` was described as unwired; it was wired on 2026-07-25 (`08c85aa`). The two-week cadence doubles the instruction lock, so that decision is explicitly reopened rather than inherited.

**Status of the plan's 12 tasks:** 1–4 valid as written · 5 blocked (premise withdrawn) · 6, 9, 10 valid, line numbers refreshed · 7 superseded · 8 done by other means, one loose end (`FooterCTA.tsx:10`) · 11 conditional · 12 rewrite.

**Merging does not start any work.** Phase 0 remains blocked on the ~$4.97 probe spend, which still needs an explicit go-ahead. Allan's sign-off on the `01/02/03` storyline is still outstanding — note #357 already renumbered it without asking.

### DIFF
--- a/docs/superpowers/plans/2026-08-09-participatory-competition-phase-0-1.md
+++ b/docs/superpowers/plans/2026-08-09-participatory-competition-phase-0-1.md
@@ -2,7 +2,50 @@
 
 > **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
 
+> # ⚠ Status: reconciled 2026-08-15 — DO NOT EXECUTE TASKS 5, 7 OR 8 AS WRITTEN
+>
+> This plan was written on 2026-08-09. Six days later PR #352 (`7db504d`) and
+> PR #357 (`60fa01f`) shipped a leaderboard redesign and a leaderboard-first
+> landing page. Neither was written against this plan, and between them they
+> delivered part of Phase 1 by other means — in one case in the **opposite**
+> direction.
+>
+> **Nothing in Phase 0 has been implemented.** Verified 2026-08-15 against
+> `origin/main` @ `88c7b8c`, and across every ref in the repo: no branch anywhere
+> adds `strategy_prompt` to `llm_agent.py`.
+>
+> ## Task status
+>
+> | # | Task | Status | Evidence checked 2026-08-15 |
+> |---|---|---|---|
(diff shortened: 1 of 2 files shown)


FlyM1ss:
Merged as `d9a3828`. Docs only — nothing here starts work.

Follow-ups are already tracked, so no new issues were filed:

- **#354** — the Live Trading season engine (C8). The merged spec argues for splitting it out *ahead* of Phase 2: it needs no user entries, no attempt ledger, and does not depend on the Phase 0 probe, so the Phase 2→3 ordering constraint (which is about user entry) does not bind it. That scheduling argument belongs on #354.
- **#355** — the qualifier gate and `instruction_sha256` scope. Both still block Phase 2; merging these docs does not resolve either, and the reconciled spec says so rather than letting its older framing imply an answer.

Still gated: the ~$4.97 Phase 0 probe needs an explicit go-ahead. Phase 0 is *not* blocked by #355 — it measures whether instructions move returns, which is upstream of every question in that issue.

Authority split recorded in the merged doc: `2026-08-15-live-trading-leaderboard-ui.md` wins on the boards' UI and payload contract; this spec stays authoritative on user entry, which that one puts out of scope.

PR #329, opened 2026-08-09
fix(leaderboard): stop leaking raw exception text on public GET

### LINKED ISSUE #327: fix: public leaderboard GET returns raw exception text (leaks Neon host)
`dashboard/backend/api/routers/leaderboard.py:47-48`:

```python
except Exception as exc:
    raise HTTPException(status_code=500, detail=str(exc)) from exc
```

`GET /api/v1/leaderboard` is public and unauthenticated, so this puts arbitrary internal exception text into a response body any visitor can read.

**Concrete leak.** `get_leaderboard()` reaches `db.get_runs_by_session()`, and prod has `AGENT_RUNS_DATABASE_URL` set, so `db` is `PostgresBacktestDatabase`. A psycopg connection failure message embeds the server host — i.e. the Neon endpoint id lands in the 500 body. That is not hypothetical: first-request-after-idle Neon failures are a known mode here (#294's stale-pooled-connection fix), and this repo already treats Neon endpoint ids as sensitive enough to keep out of test fixtures.

**Why it isn't already an alert.** The sibling handler on the same router (`POST /daily/refresh`) had the identical shape and CodeQL flagged it as `py/stack-trace-exposure` inside PR #325, where it was fixed by returning a static message. The GET is pre-existing, unflagged, and outside that PR's diff — diff-scoped analysis never reaches it. Not currently in main's open or dismissed set (checked both).

**Suggested fix** — mirror what the POST sibling already does:

```python
except Exception:
    print(f"⚠️ Leaderboard request failed: {traceback.format_exc()}")
    raise HTTPException(status_code=500, detail="Failed to load leaderboard") from None
```

`print()` rather than `logger` per the repo convention (logging is invisible under deployed uvicorn), and `from None` so the chain isn't re-attached.

Worth grepping the other routers for the same `detail=str(exc)` shape while in here. Related: #193.

hudsonwa:
## Why

`GET /api/v1/leaderboard` is public and unauthenticated, but its exception handlers returned `detail=str(exc)`. Two different leaks came out of that:

- A psycopg connection failure message embeds the database server host — in prod that's the Neon endpoint id — reachable through the generic `500` branch.
- `AlpacaCredentialsError` embeds the **server-side credentials path** (`.../credentials/alpaca.json`), reachable through the `503` branch, because `MarketDataUnavailableError` subclasses `RuntimeError`.

This is the same `py/stack-trace-exposure` shape CodeQL flagged in PR #325 (fixed there for the POST sibling); the GET was pre-existing and outside that PR's diff.

## Summary

- Neither branch echoes exception text any more. Both log the traceback server-side (`print(f"...{traceback.format_exc()}")`, per repo convention) and return a static message: `"Failed to load leaderboard"` (500) / `"Leaderboard is temporarily unavailable"` (503).
- `RuntimeError` now selects only the **status code**, never the message. It is not an allowlist of safe domain errors: it is the base class of this repo's market-data family — `MarketDataUnavailableError` → `AlpacaCredentialsError`, `MarketDataDependencyError`, `MarketDataCredentialsError`, `IFindClientError` — all of which reach this handler via `get_leaderboard` → `ensure_leaderboard_runs` → `fetch_hourly_bars` → `AlpacaDataLoader()`. Those classes subclass `RuntimeError` to escape `SystemExit` (see the `alpaca_bars` module docstring), not as a safety marker.
- Two regression tests: one for each branch. The 503 test raises through the *real* `AlpacaDataLoader` rather than a hand-written message, so a reworded upstream string cannot drift away from the assertion.

## Issue Number

Fixes Open-Finance-Lab/AgenticTrading#327

## How to Test

1. `pytest dashboard/backend/tests/test_leaderboard_api.py` — 23 passed (incl. both leak regressions)
2. Full suite: `pytest dashboard/backend/tests/` — 2653 passed, 67 skipped

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The traceback still reaches operators via stdout (uvicorn logs), so the fix only hides the leak from anonymous callers, not from the server logs.

The 503 message was previously described here as carrying deliberate domain failures ("leaderboard not ready yet") that were safe to show. That message does not exist anywhere in the codebase, and no test or frontend code reads the 503 `detail`. The `RuntimeError`s that *can* actually reach this handler are the market-data family above and `LeaderboardFallbackError` (operator-facing text about `--allow-fallback`, raised unwrapped from `ensure_leaderboard_runs`) — none of which is useful to an anonymous visitor.

Other `detail=str(exc)` sites exist in authenticated routers (backtests, agents, etc.); those are not the same exposure class since they require auth, so they're out of scope here.

### DIFF
--- a/dashboard/backend/api/routers/leaderboard.py
+++ b/dashboard/backend/api/routers/leaderboard.py
@@ -7,6 +7,8 @@ status codes, exception messages, and service calls are unchanged; only the
 module location moved.
 """
 
+import traceback
+
 from fastapi import APIRouter, Header, HTTPException, Query, Request
 from fastapi.responses import JSONResponse
 
@@ -43,9 +45,17 @@ def api_get_leaderboard(
     try:
         return get_leaderboard(force_refresh=refresh, period=period)
     except RuntimeError as exc:
+        # Deliberate domain failure (e.g. "leaderboard not ready yet") — the
+        # message is safe to show.
         raise HTTPException(status_code=503, detail=str(exc)) from exc
-    except Exception as exc:
-        raise HTTPException(status_code=500, detail=str(exc)) from exc
+    except Exception:
+        # Public, unauthenticated endpoint: a raw exception string can embed
+        # internal infrastructure details (e.g. the Neon endpoint id from a
+        # psycopg connection failure). Log the traceback server-side and
+        # answer with a static message, mirroring the POST /daily/refresh
+        # sibling fix (PR #325).
+        print(f"⚠️ Leaderboard request failed: {traceback.format_exc()}")
+        raise HTTPException(status_code=500, detail="Failed to load leaderboard") from None
 
 
 @router.post("/daily/refresh")
--- a/dashboard/backend/api/routers/leaderboard.py
+++ b/dashboard/backend/api/routers/leaderboard.py
@@ -42,18 +42,27 @@ def api_get_leaderboard(
     Pass ?refresh=true to recompute (e.g. after config change).
     Pass ?period=daily for the rolling one-day board (weekends show Friday).
     """
+    # Public, unauthenticated endpoint: no exception text reaches the caller.
+    # A raw exception string can embed internal infrastructure details — the
+    # Neon endpoint id from a psycopg failure, or the server-side credentials
+    # path from AlpacaCredentialsError. Both branches log the traceback
+    # server-side and answer with a static message, mirroring the
+    # POST /daily/refresh sibling fix (PR #325).
+    #
+    # RuntimeError is NOT an allowlist of safe domain errors: it is the base
+    # class of this repo's market-data family (MarketDataUnavailableError ->
+    # AlpacaCredentialsError, MarketDataDependencyError,
+    # MarketDataCredentialsError, IFindClientError), all of which reach here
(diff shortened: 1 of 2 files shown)


FlyM1ss:
Reviewed and pushed `d058086` to this branch (fork, maintainer edits on).

The `503` branch was still leaking. `MarketDataUnavailableError` → `AlpacaCredentialsError` subclass `RuntimeError`, and reach this handler via `get_leaderboard` → `ensure_leaderboard_runs` → `fetch_hourly_bars` → `AlpacaDataLoader()`. Probed against the old head:

```
GET /api/v1/leaderboard?period=daily   (anonymous)
503 {"detail":"Alpaca credentials not found (set ALPACA_API_KEY and
     ALPACA_SECRET_KEY, or provide /app/credentials/alpaca.json)"}
```

A server filesystem path handed to an unauthenticated visitor, on the endpoint being hardened — fires on any daily window with no cached runs yet. `RuntimeError` now selects only the status code, never the message; both branches log server-side. New test drives the real loader so a reworded upstream string can't drift away from it.

Body updated: the "leaderboard not ready yet" message it cited doesn't exist in the codebase, and nothing reads the 503 `detail`.

Full suite: 2653 passed, 67 skipped.

PR #330, opened 2026-08-09
fix(backtest): pass session_id to insert_run in CLI save path

### LINKED ISSUE #264: backtest.py CLI save raises TypeError: insert_run missing session_id
`dashboard/scripts/backtest.py:397` calls `db.insert_run(...)` without `session_id`, which `dashboard/backend/database.py:494` declares as a **required** positional parameter:

```python
def insert_run(self, run_id: str, session_id: str, agent_name: str, mode: str, ...)
```

The call passes `run_id`, `agent_name`, `mode`, dates, equity, and metrics — but never `session_id`. Every save through this path raises:

```
TypeError: insert_run() missing 1 required positional argument: 'session_id'
```

Unhandled, so `python dashboard/scripts/backtest.py` fails at the save step on **every** run.

### Why it looks fixed

The sibling `dashboard/scripts/backtest_custom_algo.py:325` passes `session_id=self.session_id` correctly, so grepping for the pattern suggests the call sites are fine.

This is the same defect class as the `/paper/start-session` bug already fixed on `main` (`api/routers/paper_trading.py:278` now passes `session_id=run_id`) — this CLI sibling was missed when that one was fixed.

### Notes

- Verified at source against `main` @ 9d46dc7 (2026-07-30).
- Surfaced by the pass-2 audit of `docs/superpowers/specs/2026-07-20-run-history-persistence-design.md`; filed as a follow-up to #262.
- Fix needs a decision on where the CLI's `session_id` comes from — the custom-algo sibling carries one on the runner object, which this script has no equivalent of.

hudsonwa:
## Why

`dashboard/scripts/backtest.py` calls `db.insert_run(...)` without the required `session_id` positional arg, so `python dashboard/scripts/backtest.py` fails at the save step on every run with `TypeError: insert_run() missing 1 required positional argument: 'session_id'`.

Unblocking that save is necessary but not sufficient: the call also had the wrong **scale** and the wrong **grouping key**, so a working save would have written silently-bad rows instead of crashing.

## Summary

- `save_backtest_to_database()` now passes `session_id="legacy-demo-session"` — the shared key every other sessionless writer uses (`BacktestEngine`'s default, `backtest_hourly_agent`'s `--session-id` default, and what `database.py` backfills onto pre-session rows). A per-run session id would mint one orphan session per invocation, unreachable by every session-scoped query (`get_runs_by_session`, `get_runs_by_sessions`, `get_run_with_session`) and by stale-run sweeps.
- `total_return` / `max_drawdown` are divided by 100. `BacktestMetrics` reports **percent** (`backtest_engine` multiplies by 100), but `agent_runs` stores **fractions** — every other `insert_run` caller writes `(final - initial) / initial`. `GET /runs` is public and unfiltered, and the frontend disambiguates by magnitude (`Math.abs(x) <= 1 ? x * 100 : x`), so a genuine +0.4% run would have been stored as `0.4`, read back as a fraction, and rendered **"+40.00%"**.
- Regression tests bind each call against the **real** `BacktestDatabase.insert_run` signature via `inspect.signature(...).bind(...)`, plus a producer-side test pinning that `BacktestMetrics` really is percent-scaled.

## Issue Number

Fixes Open-Finance-Lab/AgenticTrading#264

## How to Test

1. `pytest dashboard/backend/tests/test_backtest_cli_save.py` — 4 passed
2. Full suite: `pytest dashboard/backend/tests/` — 2655 passed, 67 skipped

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The tests import `dashboard/scripts/backtest.py` via `importlib` (it's not a package) and monkeypatch the module's `db` with a recording stub, so no real database is touched and the committed seed DB is unaffected.

The stub deliberately does **not** accept `**kwargs`. A permissive stub cannot fail the way production fails: add another required parameter to `insert_run` and the CLI would break with the identical `TypeError` while the test stayed green — precisely the defect class this file exists to catch. Binding against the real signature was verified to red against pre-PR `main` with that same `TypeError`.

An earlier revision of this PR used `session_id=run_id`, citing `/paper/start-session` as precedent. That precedent is `mode="paper"`, which surfaces through mode-scoped and global listings; backtest-mode rows surface session-scoped, so the analogy does not carry over.

### DIFF
--- a/dashboard/backend/tests/test_backtest_cli_save.py
+++ b/dashboard/backend/tests/test_backtest_cli_save.py
@@ -1,0 +1,69 @@
+"""Tests for the backtest CLI save path (dashboard/scripts/backtest.py).
+
+Regression for the "insert_run missing session_id" TypeError: the CLI's
+save_backtest_to_database() called db.insert_run() without the required
+session_id positional arg, so every `python dashboard/scripts/backtest.py`
+save raised. Each standalone CLI run has no session concept, so it now passes
+session_id=run_id — the same pattern the /paper/start-session fix uses.
+"""
+
+import importlib.util
+import sys
+from pathlib import Path
+
+import pytest
+
+_SCRIPTS_DIR = Path(__file__).resolve().parents[2] / "scripts"
+
+
+def _load_backtest_script():
+    """Import dashboard/scripts/backtest.py (not a package) in isolation."""
+    path = _SCRIPTS_DIR / "backtest.py"
+    spec = importlib.util.spec_from_file_location("backtest_script", path)
+    assert spec and spec.loader
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[spec.name] = module
+    # The script imports sibling modules (backtest_engine, _bootstrap) by bare
+    # name, so the scripts dir must be on sys.path for those to resolve.
+    sys.path.insert(0, str(_SCRIPTS_DIR))
+    try:
+        spec.loader.exec_module(module)
+    finally:
+        sys.path.remove(str(_SCRIPTS_DIR))
+    return module
+
+
+def test_save_backtest_to_database_passes_session_id(monkeypatch):
+    """insert_run must receive session_id (= run_id) so the CLI save succeeds."""
+    module = _load_backtest_script()
+
+    calls = {}
+
+    class FakeDb:
+        def insert_run(self, **kwargs):
+            calls["insert_run"] = kwargs
+
+        def insert_equity_points(self, run_id, curve):
+            calls["insert_equity_points"] = (run_id, curve)
+
+    monkeypatch.setattr(module, "db", FakeDb())
+
+    results = {
+        "agent_a": {
+            "metrics": {
+                "total_return": 0.05,
+                "sharpe_ratio": 1.2,
+                "max_drawdown": -0.03,
+                "num_trades": 5,
+            },
+            "equity_curve": [{"timestamp": "t", "equity": 100000}],
+        },
+    }
+
+    module.save_backtest_to_database(results, "2026-01-01", "2026-02-01")
+
+    kwargs = calls["insert_run"]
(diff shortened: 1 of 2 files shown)


FlyM1ss:
Reviewed and pushed `1c3c2f4` to this branch (fork, maintainer edits on). The `session_id` fix is right; unblocking the save exposed two silent-bad-data problems behind it.

**Scale.** `BacktestMetrics.total_return` / `max_drawdown` are percent (`backtest_engine.py:213,231` multiply by 100); `agent_runs` stores fractions (`engine.py:1086`, `leaderboard/baselines.py:70`, `baselines/paper.py:277`). Before this PR the call raised before writing — now percent lands in the table. `GET /runs` is public and unfiltered, and the frontend disambiguates by magnitude (`app.js:7288`, `:828`, `:4271` — `Math.abs(x) <= 1 ? x*100 : x`), so a +0.4% run stores as `0.4` and renders **"+40.00%"**. Now `/100` at the call site. (The v2 leaderboard is unaffected — `_is_v2_run` requires a `run_` prefix.)

**Grouping key.** `session_id=run_id` mints an orphan session per invocation. Sessionless writers share `legacy-demo-session` (`engine.py:150`, `backtest_hourly_agent.py:193`, the `database.py:252` backfill). The `/paper/start-session` precedent is `mode="paper"`, which surfaces through mode-scoped listings; backtest-mode rows surface session-scoped.

**Test.** `FakeDb.insert_run(self, **kwargs)` couldn't catch a recurrence of its own defect class — add another required param to `insert_run` and the CLI breaks identically while the test stays green. Now binds against the real signature; verified to red against pre-PR `main` with that same `TypeError`. Also clears CodeQL alert #1221 (`py/unused-import`).

Full suite: 2655 passed, 67 skipped.

Issue #139, opened 2026-07-17
test_plot_png_cached_per_run fails intermittently


FlyM1ss:
`dashboard/backend/tests/test_backtests_router.py::test_plot_png_cached_per_run` failed once in a full-suite run and has not reproduced since.

**Observed:** one full-suite run → `1 failed, 1057 passed, 13 skipped in 86.47s`. Three subsequent full-suite runs at the same commit → `1064 passed, 0 failed` (43-45s each). The test passes 5/5 in isolation.

**The traceback is lost.** That run was piped through `| tail -15`, which discarded the `=== FAILURES ===` block, so **nobody has seen which assertion failed.** It is unreproduced, not proven benign.

**Ruled out:**
- *Test order.* pytest's collection order is deterministic, so identical commands cannot differ by order. "Passes in isolation" does **not** imply shared state here.
- *pytest-timeout.* Installed, but no default timeout is configured and `--timeout` wasn't passed; the shell `timeout 180` didn't fire at 86s.

**Leading (unproven) hypothesis: machine contention.** The failing run took 86.47s against a normal 43-45s — ~2× slower on the same box and commit. Plausible vectors: matplotlib rendering under memory pressure, or a leftover FastAPI `TestClient` threadpool thread from an earlier test racing this test's `cache_clear()` (`_render_run_plot_png` is a module-level `@lru_cache(maxsize=128)`, i.e. shared mutable state, and sync endpoints run in a threadpool).

**Not related to #134** — that branch touches no backtest, plotting, or DB code, and this test file is untouched by it.

If it recurs, capture the full output (`pytest ... > out.txt 2>&1`, never through `tail`) and post the `FAILURES` block — that alone would probably settle it.

PR #331, opened 2026-08-09
test(backtests): make plot-png cache test hermetic with unique key

hudsonwa:
## Why

`test_plot_png_cached_per_run` failed once in a full-suite run (86s vs normal 43-45s) and never reproduced — but the traceback was lost, so nobody knows which assertion failed. The module-level `@lru_cache` on `_render_run_plot_png` and the process-wide `bt.db` singleton are real shared-state seams across the test process.

## Summary

- The test used the fixed cache key `"run_x"` on a cache shared across the whole test process. Review found no code path that can actually populate that key from outside the test (the production route is only reached via `/runs/{run_id}/plot.png` URLs no test requests), so the originally hypothesized cache-poisoning mechanism cannot be #139's cause — the real cause remains unconfirmed.
- Hardening, covering both failure directions (defense in depth, not a proven-root-cause fix):
  - unique per-invocation key (`f"run_x_{uuid.uuid4().hex}"`) so no other code path can ever collide with this test's cache entry (count 0 direction);
  - counters gated on `rid == run_id` — `bt.db` is a process-wide singleton, so while it is monkeypatched *any* thread's `db.get_run` call hits the fake; unconditional counting meant a stray call from a leftover threadpool thread could tick the count to 2 (the opposite failure direction). Same gating applied to `test_plot_png_missing_run_not_cached`.
- No production code touched.

## Issue Number

Mitigates hypothesized cross-test interference vectors of Open-Finance-Lab/AgenticTrading#139. Deliberately **not** "Fixes": the original traceback was lost and the root cause is unconfirmed, so #139 should stay open until the flake recurs (now with counters that can't be polluted) or is positively diagnosed.

## How to Test

1. `pytest dashboard/backend/tests/test_backtests_router.py -k plot_png` — passes repeatedly
2. Full router file: `pytest dashboard/backend/tests/test_backtests_router.py` — 66 passed (re-verified at `0129d78`)

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This makes the test hermetic against cross-test cache collisions *and* counter pollution, but it does not claim to have reproduced or root-caused #139 — the comment in the test says so explicitly, so the next intermittent failure gets triaged against honest documentation.

### DIFF
--- a/dashboard/backend/tests/test_backtests_router.py
+++ b/dashboard/backend/tests/test_backtests_router.py
@@ -48,20 +48,24 @@ def test_plot_png_cached_per_run(monkeypatch):
         "baseline_buyhold_run_id": None, "baseline_djia_run_id": None,
     }
 
-    # Unique per-invocation key: the render cache is a module-level lru_cache
-    # shared across tests, and a leftover TestClient threadpool render (from an
-    # earlier test) that lands *after* cache_clear() would populate a fixed
-    # "run_x" entry with real data — making the 2nd call a cache hit and the
-    # get_run count 0 instead of 1 (intermittent flake). A unique key keeps
-    # this test hermetic against that race.
+    # #139's traceback was lost, so the real flake mechanism is unconfirmed.
+    # Defense in depth against cross-test interference: the render cache is a
+    # module-level lru_cache and bt.db is a process-wide singleton (patched
+    # for any thread while this test runs), so (a) a unique per-invocation
+    # key guarantees no other code path can collide with this test's cache
+    # entry, and (b) the counters only count calls for *this* key, so a stray
+    # db.get_run from another test's leftover threadpool thread cannot skew
+    # the assertions either way.
     run_id = f"run_x_{uuid.uuid4().hex}"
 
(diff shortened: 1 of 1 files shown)


FlyM1ss:
Reviewed and pushed `0129d78`:

1. **The documented flake mechanism was impossible** — repo-wide, nothing outside this test ever renders a `run_x*` cache key (the production route is only reachable via `/runs/{run_id}/plot.png`, which no test requests), so a leftover threadpool render could never have populated the old fixed key. Test comment rewritten; since #139's traceback was lost and the root cause is unconfirmed, the PR body no longer says "Fixes" (issue should stay open).
2. **The counters weren't hermetic** — `bt.db` is a process-wide singleton, so the monkeypatched fakes counted calls for *any* rid; the very straggler-thread race this PR guards against could still tick `calls["get_run"]` to 2 and fail the `== 1` assertion. Counters are now gated on the test's own key (applied to both plot-png tests).

66/66 pass in `test_backtests_router.py` at `0129d78`.

Issue #298, opened 2026-08-03
perf: threadpool 40-thread limiter is now the throughput ceiling


FlyM1ss:
Follow-up from #292. Moving blocking handlers off the event loop was the right fix, but it relocates the bottleneck rather than removing it.

Starlette dispatches plain `def` handlers through anyio, whose default limiter is **40 threads** (verified: `anyio.to_thread.current_default_thread_limiter().total_tokens == 40`). Once all 40 are busy, further requests queue — the server stays responsive but throughput caps.

Handlers that can hold a thread for a long time:

- `POST /api/algo/chat` → synchronous `anthropic.messages.create()`, i.e. seconds to a minute per call.
- `POST /backtest/run` → runs a backtest in-request.
- any `/paper/*` route when Alpaca is slow (10s client timeout).

This matters against the 100-agent scale target, where throughput was already measured collapsing 104 → 26.5 req/s.

Worth considering:
- Raise the limiter deliberately (`anyio.to_thread.current_default_thread_limiter().total_tokens = N`) rather than inheriting 40 by accident.
- Move genuinely long work (`algo_chat`, `backtest/run`) onto a job/queue instead of holding a request thread — `execute_algo` already returns a job id and polls, so `algo_chat` is the odd one out.
- No alarm currently fires when the pool saturates; it looks like ordinary latency.

PR #332, opened 2026-08-09
perf(auth): move auth.py handlers off the event loop

### LINKED ISSUE #297: perf: auth.py runs bcrypt on the event loop
Follow-up from #292, which moved 58 blocking handlers to the threadpool but left `auth.py`.

These `async def` handlers do bcrypt work directly on the event loop, so each one stalls every concurrent request for the duration of the hash (bcrypt is deliberately slow — tens to hundreds of ms):

| handler | has `await`? | password work |
|---|---|---|
| `change_password` | **no** | `verify_password` |
| `signup` | yes | `bcrypt` |
| `login` | yes | `verify_password` + `bcrypt` |
| `request_email_change` | yes | `verify_password` |

`change_password` contains no `await` at all, so it can simply become a plain `def` like the handlers in #292. The other three genuinely await, so they need the bcrypt call wrapped in `run_in_threadpool` rather than a signature change.

`login` is the one that matters most: it is on the boot path and is rate-limited per-IP/per-email (#281), so it is also the easiest to hit repeatedly.

Guard: `dashboard/backend/tests/test_event_loop_threadpool.py` already has a `BLOCKING_IO_HANDLERS` set for pinning individual handlers in modules that legitimately mix awaited and blocking work — add them there.

hudsonwa:
Closes #297. Follow-up to #292, which moved 58 blocking handlers to the threadpool but left `auth.py`.

**What was wrong:** `auth.py`'s handlers were `async def` while doing synchronous work on the event loop, so each request stalled every concurrent one — bcrypt for ~190 ms, plus store round trips that are network calls against Neon in prod.

**Changes**

- `change_password` → plain `def`. It contains no `await`, so per #297 the signature change is the fix: it moves the verify, the bcrypt **hash** inside `update_password()`, and three store round trips to the threadpool. Wrapping only the verify in `asyncio.to_thread` would have left an identically slow hash on the loop. Password-policy check hoisted above the bcrypt, matching `signup()` — a rejected new password now costs no hash, at the price of reporting the policy error first when a request is wrong on both counts.
- `request_email_change` stays `async def` (it awaits `send_email`), so its blocking half — the verify plus four store round trips — moves in one `to_thread` hop, and the insert in another.
- The other eight no-`await` `async def` handlers in `auth.py` (`me`, `logout_all`, `update_display_name`, `get_email_change`, `cancel_email_change`, `set_avatar`, `delete_avatar`, `discord_oauth_start`) flipped to `def` as well. Every route handler in the file is now either `async` with real awaits or plain `def` with none.

**Guard:** all ten pinned in `BLOCKING_IO_HANDLERS` (`tests/test_event_loop_threadpool.py`), as #297 asks. A pinned handler is only covered while it has no `await`, so adding one to offload a single call now fails the guard rather than silently exempting the handler.

**Tests:** `test_password_verify_never_runs_on_the_event_loop` asserts the property — `asyncio.get_running_loop()` raises on the thread doing the hash — rather than which dispatcher was used, so it survives a refactor to `run_in_threadpool` or a `def` flip and goes red only when the hash is back on the loop. Covers both routes. Mutation-checked: putting either handler back on the loop yields `[True, True]`. Suite 2652 passed, 67 skipped.

`signup`/`login` already offload via `asyncio.to_thread` (added in #281, not #292) — #297's table rows for those two are stale. Pool sizing across the two threadpools is #298's scope, not this PR's.

### DIFF
--- a/dashboard/backend/api/auth.py
+++ b/dashboard/backend/api/auth.py
@@ -478,7 +478,12 @@ async def change_password(
     current_user: dict = Depends(get_current_user),
     authorization: Optional[str] = Header(default=None),
 ):
-    if not verify_password(payload.current_password, current_user["password_hash"]):
+    # bcrypt verify off the event loop: it is deliberately slow (~190 ms),
+    # and on a shared event loop it would stall every concurrent request
+    # (same reason login()/signup() already offload via asyncio.to_thread).
+    if not await asyncio.to_thread(
+        verify_password, payload.current_password, current_user["password_hash"]
+    ):
         raise HTTPException(status_code=400, detail="Current password is incorrect.")
     violations = validate_new_password(payload.new_password, current_user["email"])
     if violations:
@@ -608,7 +613,11 @@ async def request_email_change(
     current_user: dict = Depends(get_current_user),
 ):
     store = users_module.user_store
-    if not verify_password(payload.current_password, current_user["password_hash"]):
+    # bcrypt verify off the event loop (same as change_password): it is
+    # deliberately slow and would stall concurrent requests when inline.
+    if not await asyncio.to_thread(
+        verify_password, payload.current_password, current_user["password_hash"]
+    ):
         raise HTTPException(status_code=400, detail="Current password is incorrect.")
     if payload.new_email == str(current_user["email"]).strip().lower():
         raise HTTPException(status_code=400, detail="That is already your email address.")
--- a/dashboard/backend/api/auth.py
+++ b/dashboard/backend/api/auth.py
@@ -427,7 +427,7 @@ async def login(payload: LoginRequest, request: Request):
 
 
 @router.get("/me")
-async def me(
+def me(
     request: Request,
     authorization: Optional[str] = Header(default=None),
     current_user: dict = Depends(get_current_user),
@@ -461,7 +461,7 @@ def logout(
 
 
 @router.post("/logout-all")
-async def logout_all(request: Request, current_user: dict = Depends(get_current_user)):
+def logout_all(request: Request, current_user: dict = Depends(get_current_user)):
     users_module.user_store.delete_other_sessions(current_user["id"], keep_token=None)
     response = JSONResponse({"status": "ok"})
     clear_session_cookie(response)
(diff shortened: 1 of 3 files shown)


FlyM1ss:
Reviewed and pushed fixes to this branch (`a6ae913`, `26fcb13`). Body updated too — it said "the bcrypt cost is entirely in the verify", which `users.py:571` / `users_postgres.py:351` contradict.

**The main correction.** `update_password()` calls `hash_password()` inline, so a successful change-password was still paying a full bcrypt round (~190 ms) on the event loop — the offload covered half the cost. And #297 prescribes a plain `def` for this handler *precisely because* it has no `await`; adding one also silently took it out of the `BLOCKING_IO_HANDLERS` guard, which is how it escaped #292 in the first place.

- `change_password` → plain `def`. Moves the verify, the hash, and three store round trips to the threadpool. Policy check hoisted above the bcrypt, matching `signup()`.
- `request_email_change` keeps `async def` for `send_email`, but its blocking half — verify plus four store round trips, which are Neon calls in prod — now moves in one `to_thread` hop.
- The other eight no-`await` `async def` handlers in `auth.py` flipped to `def`. Every route handler in the file is now either `async` with real awaits or plain `def` with none.
- All ten pinned in `BLOCKING_IO_HANDLERS`, per #297's guard note.
- Regression test asserts the property — `asyncio.get_running_loop()` raises on the thread doing the hash — rather than which dispatcher was called, so it survives a refactor and fails only when the hash is back on the loop. Covers both routes; no longer patches `asyncio.to_thread` on the stdlib module process-wide.

Kept `asyncio.to_thread` rather than `run_in_threadpool`: measured 16 workers vs anyio's 40, the file has nine existing `to_thread` sites (the pattern came from #281, not #292), and the heaviest bcrypt path is on the anyio side now anyway. Pool sizing is #298's scope.

Verified: mutating either handler back onto the loop turns the test red with `[True, True]`; CI 2703 passed / 7 skipped (skip collapse from 67 confirms the `@pg_only` tier ran); CodeQL `#1222` (`py/import-and-import-from`) cleared, 0 open alerts on `refs/pull/332/merge`.

PR #335, opened 2026-08-11
chore(leaderboard): temporarily hide Daily Leaderboard

JinBoatus1:
## Summary
- Temporarily hide the Daily Leaderboard tab (and About blurb); `?view=daily` redirects to Competition Leaderboard
- Pause the nightly refresh cron; keep backend + manual `workflow_dispatch`
- Backend/API and leaderboard.js left in place for a later re-enable

## Test plan
- [ ] Competition page shows no Daily Leaderboard subtab
- [ ] Opening `?view=daily` lands on Competition Leaderboard
- [ ] Contest board still loads normally
- [ ] `pytest dashboard/backend/tests/test_frontend_daily_leaderboard.py -v`


FlyM1ss:
Heads-up from #348 — this PR's cache busters need raising before merge.

`main` currently serves `app.js?v=88` / `styles.css?v=87`; this branch ships `78` / `86`. Both are numbers that already shipped, so returning browsers would keep running their cached bundle and never pick up the `app.js` change here.

That's also why **Backend tests** is red: `test_cache_busters_bumped` still asserts `app.js?v=77` from this branch's base. Worth knowing that the guard is only checking `app.html` against the test file — it does *not* compare against `main` (#347), so setting the assert to `78` would turn it green while leaving the stale-cache problem in place.

Fix is to pick a number above whatever `main` serves at merge time in **both** `app.html` and the guard. #341 is currently sitting on `98`/`97`, so re-check the floor rather than hardcoding off this comment.

FlyM1ss:
Heads-up on an overlap, not a request to change anything here.

#352 (draft) replaces the Daily Leaderboard tab with a two-week Live Season board, following the design Felix and @JinBoatus1 discussed. It touches every region this PR touches — the `daily:` nav-map entry, `migrateSavedNavState`, the subtab button, the About card, `viewParamForNavState`, `showCompetitionPanel` — plus it renames `test_frontend_daily_leaderboard.py`, which this PR adds to. So they conflict rather than compose, and whichever lands second needs a rebase.

The cron pause from this PR is carried over in #352 (c53e598), since a board with no tab should not be deploying models nightly either way.

Full comparison: https://github.com/Open-Finance-Lab/AgenticTrading/pull/352#issuecomment-5297501238 — ordering is yours and Felix's call.

PR #338, opened 2026-08-12
feat(backtest): enforce A-share daily market rules

MrParamecium:
## Summary
- load official iFinD daily trading status, closing limit status, and unadjusted close data for registered A-share universes
- enforce full-day suspensions and direction-aware closing limit gates for Agent and buy-and-hold simulation
- persist and display native-CNY rule audits, rejection totals, and delayed or unfilled baseline allocations

## Data and execution integrity
- use unadjusted iFinD prices (`CPS=no` for high-frequency data and `CPS=1` for daily data)
- supplement only blank official status fields through same-date basic data
- fail before execution when market-rule coverage or final-bar close alignment is invalid
- do not infer historical intraday limit touches or theoretical limit prices
- leave Alpaca and vn.py paths unchanged

## Scope
ATL remains a backtest and paper-simulation platform. This change does not place live orders.

## Validation
- `python -m pytest dashboard/backend/tests -q`: 2729 passed, 66 skipped
- `node --check dashboard/frontend/app.js`
- `git diff --check`
- desktop and 390x844 browser QA with offline fixtures

### DIFF
--- a/dashboard/backend/api/routers/backtests.py
+++ b/dashboard/backend/api/routers/backtests.py
@@ -260,6 +260,8 @@ class RunMetadata(BaseModel):
     # curve places no orders, so it carries the profile with the flag false.
     transaction_costs_applied: Optional[bool] = None
     transaction_cost_totals: Optional[Dict[str, Any]] = None
+    market_rule_profile: Optional[Dict[str, Any]] = None
+    market_rule_rejections: Optional[Dict[str, int]] = None
     # How much of the buy & hold sleeve filled. A lot-constrained benchmark on
     # a small account can place fewer symbols than requested, and a partly
     # placed benchmark must be legible rather than read as a real flat curve.
@@ -341,6 +343,8 @@ def _run_metadata_response(run: Dict[str, Any]) -> RunMetadata:
             "transaction_cost_profile",
             "transaction_costs_applied",
             "transaction_cost_totals",
+            "market_rule_profile",
+            "market_rule_rejections",
             "baseline_allocation",
             "rejected_orders_count",
             "rejected_orders_truncated",
--- a/dashboard/backend/baseline_generator.py
+++ b/dashboard/backend/baseline_generator.py
@@ -21,6 +21,7 @@ from datetime import datetime, time
 from dashboard.backend.paths import CREDENTIALS_DIR
 from dashboard.backend.domain.backtesting.constants import INITIAL_CAPITAL
 from dashboard.backend.domain.backtesting.currency import CurrencyContext
+from dashboard.backend.domain.backtesting.market_rules import MarketRuleCalendar
 from dashboard.backend.domain.trading.execution import calculate_transaction_costs
 from dashboard.backend.infrastructure.market_data.alpaca_bars import (
     MarketDataUnavailableError,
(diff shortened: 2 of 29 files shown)


FlyM1ss:
Reviewed and pushed 3 fix commits to this branch (`e257169`, `abab9a2`, `5909609`) — 7 review findings plus one found while fixing. Each has a regression test that fails against the prior code.

| Fix | Why |
| --- | --- |
| `baseline_generator.py` — `reserved_capital` withheld from the top-up sweep | Pass 2 budgeted on `capital - spent` over the *filtered* price map, so it spent a blocked symbol's slice on the others. The deferred retry then had ¥875 instead of ¥50k and could never fill. This one silently corrupts benchmark curves. |
| `market_rules.py` + `ifind_market_rules.py` — active symbol-dates with no hourly bars are kept ungated (`final_bar_timestamp=None`) | The union-of-dates load aborted the whole run whenever any symbol lacked bars on another symbol's trading date. Own-date close alignment is unchanged, so a wholesale feed break still fails loudly. |
| `ifind_market_rules.py` — failures name the value, symbol and date; status vs. missing-close split | Exact-string failures were undiagnosable. Deliberately **no** permissive fallback: guessing at an unrecognised status is how a limit-locked day gets traded. |
| `ifind_market_rules.py` — skip the supplement round trip when the row already proves suspension | Saved one blocking vendor call per suspended symbol-date. |
| `execution.py` — a symbol outside the calendar is rejected (`market_rule_unavailable`), not raised | A missing rule killed the run mid-backtest. Now fail-closed and recorded in the audit. |
| `app.js` — audit line only on rows a rule spoke to | It printed on every ordinary fill. |
| `app.html` — `app.js?v=77` → `78` | The PR changed `app.js` without bumping the buster, so returning browsers would have kept the cached bundle and never seen any of the new UI. |

**Known limitation, not fixed:** `CPS=no`/`CPS=1` are unadjusted on purpose (the close-alignment check needs like-for-like), so 除权除息 dates land in the equity curve and the buy-and-hold baseline as a loss that never happened. Real fix is corporate-action handling — out of scope here. Mitigated with a >21% overnight-gap warning and written up in the design doc.

CI green, CodeQL 0 on the merge ref. Suite: 2732 passed, 67 skipped.

FlyM1ss:
Merged `main` in (`50e50dd`) — #339 landed while this was in flight and both PRs bumped the same cache buster.

Resolved to `app.js?v=88` rather than either side's value: #339's `v=87` is already deployed, so taking `77`/`78` would have served a stale bundle and taking `87` would have left returning browsers on #339's cached copy without the market-rule UI. Both sides' `app.js` changes verified present after the auto-merge. Suite green on the merged tree: 2744 passed, 69 skipped.

FlyM1ss:
Merged as `cec0a89`; prod deploy hook fired green.

The one finding not fixed here is now tracked as #346 — unadjusted prices making a 除权除息 date read as a real loss. Worth noting it is a **behaviour change**, not a pre-existing gap: the hourly feed was `CPS=forward1` before this PR (`ece7c6d` changed it), which smoothed the ex-rights drop away. The unadjusted request is still the right call for the rule audit and the close-alignment check — #346 covers doing both properly rather than reverting.

PR #339, opened 2026-08-13
fix(agents): isolate account-bound agents across login/logout

Allan-Feng:
## Summary
- Anonymous browser agent listing no longer returns agents already bound to another account.
- `owns_agent` / `claim` / `reclaim` refuse cross-account takeover on a shared browser.
- Logout clears the active agent trading session so the next login cannot inherit it.

## Why
Same-browser account switching (and logout) could surface another user's agents, and activate/claim could reassign `owner_user_id`. This is live on `main` today.

## Test plan
- [x] `pytest dashboard/backend/tests/domain/agents/test_repository_move.py dashboard/backend/tests/domain/agents/test_service.py dashboard/backend/tests/test_agents_api.py -q`
- [ ] On one browser: create agents while logged in as A → logout → list should not show A's agents
- [ ] Log in as B on the same browser → B must not see or activate A's agents

### DIFF
--- a/dashboard/backend/api/dependencies.py
+++ b/dashboard/backend/api/dependencies.py
@@ -66,15 +66,38 @@ def _require_agent_access(
     api_key: Optional[str] = None,
     reclaim_on_session_match: bool = False,
 ) -> Dict[str, Any]:
+    agent, _owned = _resolve_agent_access(
+        agent_id,
+        ctx,
+        api_key=api_key,
+        reclaim_on_session_match=reclaim_on_session_match,
+    )
+    return agent
+
+
+def _resolve_agent_access(
+    agent_id: str,
+    ctx: Dict[str, Any],
+    *,
+    api_key: Optional[str] = None,
+    reclaim_on_session_match: bool = False,
+) -> tuple[Dict[str, Any], bool]:
+    """``(agent, owned)``. See ``AgentService.resolve_access``.
+
+    Routes that only *read* or edit the agent can use ``_require_agent_access``
+    and ignore the flag. Routes that write ownership must not: a grant that came
+    from a bare ``session_id`` match is not proof of ownership, and binding an
+    account on it cannot be undone by the real owner.
+    """
     # The agent's own API key is a valid credential for its own agent — this is
     # how API-only clients (which have no logged-in user or browser session)
     # authorize state-changing operations on the agent they own.
     if api_key:
         keyed = agent_service.resolve_api_key(api_key)
         if keyed and keyed.get("agent_id") == agent_id:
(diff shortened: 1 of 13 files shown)


FlyM1ss:
Reviewed and pushed 7 fixes to this branch as `0318b13`.

**Blocking (would have 500'd in prod):** `claim_agent`/`reclaim_agent` in the Postgres twin bind `owner_user_id` to a bare `%s IS NOT NULL`. psycopg sends `None` with an unspecified type OID, so Postgres raises `42P08 could not determine data type of parameter $n` — on the *anonymous* path, i.e. every logged-out activate/restore. Added `::integer` casts. The SQLite twin can't reproduce it (untyped params), and changing the test from `owner_user_id=None` to `=7` removed the only input that triggers it, so the `@pg_only` coverage is restored.

**Ownership:**
- A bare `session_id` match no longer writes `owner_user_id`. It grants access and refreshes the browser stamp only — `resolve_access() -> (agent, owned)`; the activate route passes `user_id=... if owned else None`. Binding an account from a session id is irreversible (`owns_agent` then refuses browser-only access), and the id comes off a caller-supplied header.
- `register_or_get_agent` was a 4th, unguarded writer of `owner_user_id`, reachable from `POST /agents/import-session`. Guarded in both twins + 403 at the service layer.
- The anonymous list fold now accepts `owner_browser_session = session_id` (the import-session shape) — requiring a browser-id match hid an agent from its own owner.

**Frontend:** `refreshAuthUser`'s bare catch funnelled *every* `/api/auth/me` failure into `clearAuthState`, so a cold start or transient 500 wiped the active agent. Split out `clearActiveAgentSession()`, called from logout and a real 401 only (`AuthAPI.request` now sets `error.status`).

Plus: dropped the unused `count_agents` from both twins. 9 tests added; CI green, both new `@pg_only` cases confirmed executed against CI Postgres (skip count unchanged, +9 passed). CodeQL 0 open.

Not fixed, deliberately: a guest agent's `session_id` remains a reclaim credential for the browser stamp — closing that reddens `test_patch_agent_legacy_session_owner`, which pins the behaviour as intended. It's reversible again, and `session_id` is no longer exposed by the public `/agents/builtin` listing. Worth a follow-up issue if we want it closed properly.

FlyM1ss:
Follow-up filed as #345 — the residual `session_id`-as-reclaim-credential risk described above. Constrained to unclaimed rows and reversible after this PR; left open because closing it reddens `test_patch_agent_legacy_session_owner`, which pins the legacy dashboard reclaim as intended.

Merged as `0613260`; prod deploy verified serving the new image.

PR #340, opened 2026-08-13
fix(market-data): clamp SIP end for Basic Alpaca plans

Allan-Feng:
## Summary
- Default Alpaca feed to SIP for backtests/baselines.
- Clamp SIP `end` to now−15 minutes so Basic plans stay outside the forbidden recent window.
- Fall back to IEX on subscription errors; optional `ALPACA_ALLOW_RECENT_SIP=1` for paid plans.

## Why
Daily leaderboard `end_date + 1 day` often lands on “tomorrow”, which Basic SIP rejects (`subscription does not permit querying recent SIP data`).

## Test plan
- [x] `pytest dashboard/backend/tests/infrastructure/market_data/test_alpaca_bars.py -q`
- [ ] Trigger daily/hourly bar fetch near market open and confirm no SIP recent-window 403


FlyM1ss:
Reviewed this branch and pushed the fixes directly: `a515778` (9 findings) and `345f848` (the CodeQL notes those fixes themselves minted). All checks green; 0 CodeQL alerts on the merge ref.

**The largest change isn't the one in the title.** alpaca-py defaulted to IEX; we now send `feed=` explicitly with a SIP default. Every curve computed after this merge is priced off a different tape than every one before it, and `_find_cached_run` keys on `(mode, start_date, end_date, llm_model)` alone — so old IEX rows and new SIP rows rank side by side.

What the fixes cover:

- **Provenance.** Runs record `market_data_feed` / `sip_fallback_to_iex` / `end_clamped` in `agent_runs.metadata`, baselines included. `ensure_leaderboard_runs` warns once per process when cached rows disagree with the configured feed — it warns rather than auto-refreshing, because that path serves a public unauthenticated GET and mismatch-as-missing would refetch Alpaca on every page load.
- **The clamp.** Floored at `start` (a same-day run dispatched just after 00:00 UTC clamped `end` to *yesterday*, and the negative cache pinned the empty result for the whole TTL); returns an unparseable `end` unchanged rather than raising ahead of the SDK's own validation. Docstring corrected to the real semantics: Alpaca filters on the bar's **opening** timestamp, so the 15-minute default keeps the complete 15:00–16:00 ET bar. The margin is one bar wide — `ALPACA_SIP_DELAY_MINUTES` above ~65 does drop the closing hour, now pinned by a test.
- **Config.** An unknown `ALPACA_DATA_FEED` raises instead of silently becoming SIP — a typo'd `IEXX` meant to *force* IEX previously shipped the opposite. The read-only comparison path swallows it so a typo can't 500 a cached board.
- **Duplication.** `BaselineGenerator._fetch_bars_for_symbol` delegates to `AlpacaDataLoader` instead of hand-rolling a second SIP request with no IEX retry, which caught the exact subscription refusal this PR exists to absorb and returned it as `None` — "the feed refused us" rendered as "there is no data".
- **Docs.** `ALPACA_DATA_FEED`, `ALPACA_SIP_DELAY_MINUTES`, `ALPACA_ALLOW_RECENT_SIP` in `.env.example` and CLAUDE.md.

12 new tests; suite at 2706 passed / 67 skipped.

Operator note for after this deploys: cached rows predate the tape flip. Baselines recompute with `GET /api/v1/leaderboard?refresh=true`, but LLM entries only recompute through `deploy_model_run` (billable API calls), so the board wants refreshing on one tape rather than half.

PR #341, opened 2026-08-13
feat(admin): entitlements, roles, and admin console

### LINKED ISSUE #351: func(admin): credits entitlement is admin-editable but metered nowhere
Follow-up from #341.

`user_entitlements.credits` is stored, range-validated against `MAX_CREDITS_CAP`, settable from the admin console, and returned to the client on signup, login, and `/auth/me`. Nothing reads it. There is no deduction anywhere in `domain/` — every `credits` hit outside `users.py` and `admin_users.py` is a comment.

#341 labelled the admin-console column `(not enforced yet)` so an operator cannot mistake setting it for buying anything, and `check_owner_active_run_cap`'s docstring records why it was left alone: nothing in the platform defines what one credit buys, so there is no metering to enforce, and inventing one inside a security fix would have been a product decision.

That decision is the actual work here:

1. What does a credit denominate — an LLM token, a backtest hour, a run?
2. Where is it spent? `/backtest/run` burns real operator LLM budget per trading hour; `/api/v2/runs` does not.
3. What happens at zero — refuse the run, or degrade to rule-based?

Until that is answered the field is a number with a cap on it. Either implement the metering or drop the column: an admin-editable knob wired to nothing invites the next person to assume it works.

Allan-Feng:
## Summary
- Per-user entitlements: `max_concurrent_backtests` and `credits`, both enforced.
- Admin APIs: list/patch users, bootstrap, site stats (`/api/admin/*`), all behind one router-level `require_admin`.
- Admin UI in the profile menu (role-gated) with users/admins/agents/running counters.

Closes #351.

## Notes
- Signup remains `role=user`; admin is assigned via bootstrap/SQL/admin UI.
- `max_concurrent_backtests` binds on both run surfaces (v1 protocol + v2), one ledger.
- `credits` meters LLM-driven dashboard backtests, 1 per run. **Strict opt-in** — off unless `CREDITS_METERING_ENABLED` is set, so merging changes no behaviour. Arming it makes `/backtest/run` sign-in-only for LLM runs. See CLAUDE.md.
- Prod env already set: `ADMIN_BOOTSTRAP_SECRET`, `USERS_DATABASE_URL`.

## Test plan
- [x] `pytest dashboard/backend/tests/ -q` → 2826 passed, 78 skipped
- [x] `test_credit_metering.py`, `test_admin_users.py`, `test_admin_console_frontend.py`
- [ ] Postgres tier (`@pg_only`) runs on CI only — verify the credit-ledger cases in the job log
- [ ] Promote an account, open Admin, edit max concurrent / credits


FlyM1ss:
Follow-ups from this review, filed so they don't die in a comment:

- #347 — the cache-buster guard asserts this branch's own `?v=` and never compares against `main`, so it cannot see the collision it exists to prevent. This PR hit it (bumped to `88`, which `main` already served); now on `98`.
- #348 — three other open PRs (#335, #342, #344) currently ship `app.js` busters at or below `main`'s `88`.
- #349 — the bootstrap rate-limit budgets added here are per-process, so the "server-wide" ceiling is really per-worker. Not exploitable on the current single-instance deploy.

Nothing here blocks this PR.

FlyM1ss:
Round-2 review fixes pushed (`a76bc15`..`080a9eb`) — the 15 verified findings plus 7 more confirmed by an adversarial pass over the fix diff itself.

Highlights:
- **`max_concurrent_backtests` is now enforced**: an account's agents share its active-run budget on both v1 and v2 run creation (429 `too_many_active_runs_for_account`). The entitlement/ownership reads run **off** the shared create lock (no Neon latency inside it) and fail open if the users/content DB is unreachable, so run creation gains no hard dependency on those stores.
- `apply_admin_patch` applies role + entitlements in one transaction on both store twins; `set_user_role` folded into it.
- Bootstrap secret guessing bounded per-IP + a global ceiling (header rotation can't dodge it); explicit-null PATCH bodies now 400 instead of silently no-oping.
- Admin console: PATCH sends only changed fields, request-seq guard owns every table paint, quota bounds mirrored from the server caps, boot gate stops the pre-hydration admin flash.
- Coverage: per-account cap tested on v1 **and** v2; entitlements upsert/FK-mapping/atomic-patch tested under the CI Postgres tier (verified executed in the job log, not skipped).

CI green on `080a9eb`; CodeQL: 0 open alerts on the merge ref.

FlyM1ss:
Security review round pushed as `29c0ebc`. CI green across the board, 0 CodeQL alerts on the merge ref.

**Fixed**

- Two ways past the new per-account run cap: unclaimed agents were billed to nobody (so signing out bought an unlimited budget), and `/api/v1/backtest/start` — no auth, no `protocol_runs` row — was the one unbounded path into the same engine. Both closed; the legacy surface now carries its own per-session and global budgets.
- Default quota was `1`, which would have throttled every existing user the moment this deployed. Now `5`, env-overridable, and `0` is a real "suspended" value rather than a validation error.
- Bootstrap: the global attempt budget was spent *before* the secret compare, so a burst of wrong guesses could lock the operator out of their own promotion path; there was no entropy floor; and 503-for-unset vs 403-for-wrong told an anonymous caller whether the endpoint was armed. Budget now applies only after a failed compare, 32-char minimum, 403 for unset/weak/wrong alike.
- Router-level `require_admin` (rather than per-route, where the next route added is the one that forgets), an audit line on every privileged mutation, control-character rejection at the email validator, store I/O off the event loop in signup/login, and a thread-safe limiter with an atomic `allow()`.

Prod was checked over the Render API: `ADMIN_BOOTSTRAP_SECRET` is 43 chars and `USERS_DATABASE_URL` is set, so the one-shot-promotion finding was never live there and the new floor won't strand the deployed value.

**Filed rather than fixed** — #350 (the browser-session cap is an incentive fix, not a bound against a caller who rotates the header; only `MAX_ACTIVE_RUNS_GLOBAL` binds) and #351 (`credits` is admin-editable but metered nowhere — now labelled as such in the UI).

PR #342, opened 2026-08-13
feat(backtests): honor per-user concurrent dashboard slots

Allan-Feng:
## Summary
- Replace process-wide single-flight dashboard backtests with entitlement-aware slots.
- Poll every in-flight `live_run_id` so each running agent card shows its own progress.
- Keep Configure clickable while a backtest is running.

## Stacking
**Depends on #341** (`feat/admin-entitlements`). This branch includes that commit plus the concurrent follow-up. Review the latest commit for concurrent-only changes; merge #341 first, then this.

## Test plan
- [x] `pytest dashboard/backend/tests/test_admin_users.py dashboard/backend/tests/test_backtest_progress_card.py -q`
- [ ] Set max concurrent ≥ 2, start two backtests, both cards show step/percent
- [ ] Third run refused with clear message when at cap


FlyM1ss:
Heads-up from #348 — cache-buster collision on this branch, and CI cannot see it.

This PR ships `app.js?v=88`, which is **exactly** what `main` already serves. Because the line text is identical there's no merge conflict and nothing goes red, but it changes `app.js` by ~390 lines — so on deploy, every browser holding the current `v=88` would keep running the old bundle.

`styles.css?v=97` collides with #341, which is also on `97` (and on `app.js?v=98`).

The guard only compares `app.html` against the test file, never against `main` (#347), which is why this passes. Please bump `app.js` above `main`'s floor at merge time in both `app.html` and `test_frontend_fast_boot.py`, and take `styles.css` past `97` if #341 lands first.

FlyM1ss:
Merged as 38943b45 and verified live in prod (`GET /backtest/status?live_run_id=<unknown>` now 404s where it previously returned 200).

Three things were left unfixed on purpose. Filed so they don't die in this thread:

- #362 — `max_concurrent_backtests` is now enforced once per surface, so an account's real ceiling is its quota on each rather than across both. Unifying it here would have halved every existing protocol user's capacity on this deploy.
- #363 — `MAX_ACTIVE_DASHBOARD_BACKTESTS` counts over a per-process dict, so a multi-replica deploy multiplies it. Same class as #349.
- #364 — the per-owner cap is keyed on a caller-chosen header, so it closes the incentive rather than bounding the resource. Dashboard twin of #350.

None is exploitable on today's single-instance prod.

PR #344, opened 2026-08-13
feat(billing): add Stripe Test Mode Credits recharge

MrParamecium:
## Summary

- add a Stripe Test Mode-only Credits purchase flow with fixed and custom USD amounts
- add immutable Credits ledgers and payment/refund persistence for SQLite and Postgres
- add authenticated balance, ledger, Checkout status, webhook, and admin refund APIs
- add a responsive Credits & Billing page with bounded payment polling and admin partial refunds
- document local Stripe CLI setup, Sandbox testing, and the explicit Live Mode gate

## Safety and scope

- Stripe Live Mode keys are rejected
- Checkout redirects never grant Credits; only verified Stripe webhooks can change balances
- webhook replays, checkout retries, and refund retries are idempotent
- Credits are ATL platform-service units, not simulated portfolio cash, stored money, securities, or withdrawable value
- this PR does not implement Credit consumption, signup grants, referral rewards, expiration, or self-service refunds
- no Stripe secret, webhook payload, local database, or card data is committed

## Verification

- `2805 passed, 72 skipped` in the full backend suite before the final `main` sync
- `174 passed, 6 skipped` in the focused Credits/API/security/integration suite after syncing the latest `main`
- `171 passed, 2 skipped` in the packaging SDK suite after syncing the latest `main`
- desktop and 375px mobile browser QA completed
- real Stripe Sandbox acceptance completed:
  - $10 Checkout payment produced exactly one `+10.00` ledger entry
  - Stripe CLI forwarded the signed `checkout.session.completed` webhook with HTTP 200
  - a $4 admin partial refund produced exactly one `-4.00` ledger entry
  - Stripe CLI forwarded `refund.created` and `refund.updated` with HTTP 200
  - the final balance was exactly `6.00 Credits` and the order was `Partially Refunded`


FlyM1ss:
Heads-up from #348 — this branch's `app.js` cache buster is below `main`'s.

`main` serves `app.js?v=88`; this PR ships `78`, a number that already shipped. This PR does change `app.js`, so returning browsers would keep serving their cached copy. (`styles.css?v=88` is fine — it clears `main`'s `87`.)

CI is green because `test_cache_busters_bumped` was updated to assert `78` to match. That guard only checks `app.html` against the test file and never against `main` (#347), so matching it to the shipped value hides exactly this case rather than catching it.

Please raise `app.js` above whatever `main` serves at merge time in both files — #341 is currently on `98`/`97`, so re-check the floor. This PR is also showing `mergeable: false` right now, so `app.html` will need a manual resolution anyway; a buster conflict should always resolve *upward*, never to either side.

MrParamecium:
Thanks for catching this. I merged the latest `main`, resolved the `app.html` and cache-buster test conflicts upward, and preserved the Credits script load order.

The branch now ships:
- `app.js?v=99`
- `styles.css?v=98`

Local verification after the merge:
- Backend: `2873 passed, 74 skipped`
- Packaging: `171 passed, 2 skipped`
- Focused frontend/Credits: `123 passed, 6 skipped`

The fix is in `e2364db`.

FlyM1ss:
Pushed `baaedfd`: merged current `main` (this was conflicting since #352) and fixed 14 review findings. Full backend suite 3050 passed / 84 skipped; SDK 171 passed.

**Money in, no Credits** — all four returned HTTP 200 and logged nothing:
- `settle_paid_checkout` read a NULL `stripe_checkout_session_id` as a *mismatch*, so a crash between the Stripe create and `attach_checkout_session` rejected the paid webhook permanently, and the event row made it unreplayable. Now claimed, matching the idiom the PaymentIntent check next to it already used. Both twins.
- `checkout.session.async_payment_succeeded` was unhandled while its failure twin was — unreachable on `['card']`, which is why it needed fixing before someone adds a payment method.
- Every non-settling webhook outcome now prints. Nothing distinguished "no events yet" from "an upstream rename is rejecting every payment".
- `BillingConfigurationError` subclasses `ValueError`, so it hit the 422 arm and blamed the caller for an operator misconfiguration; on `/credits/balance` it was an unhandled 500. Both 503 now.

**Refunds** — I got this one wrong first: the reservation *is* retryable by design (`refund_id` derives from `client_request_id`), and `test_refund_gateway_timeout_keeps_one_retryable_reservation` caught my over-broad fix. Corrected to classify Stripe failures instead: definitive refusals (`InvalidRequestError`/`Authentication`/`Permission` — no Refund object exists) release the reservation; ambiguous ones (timeout, dropped connection) keep it, because releasing there could refund the same money twice. Separately, editing the amount after a failed refund sent the *old* amount while the UI validated the new one.

**Also**
- `credit_accounts.status == 'restricted'` was enforced only in `credits.js`; a plain fetch bypassed it. Enforced in `create_checkout`, plus `POST /api/admin/credits/accounts/{user_id}/reinstate` — restriction is applied automatically, so making it bite without a remedy would have left SQL as the only way out.
- Postgres lock-order inversion: `reserve_refund` takes order→refund, `settle_succeeded_refund` took refund→order. A second partial refund racing a settlement webhook deadlocks (40P01) and escapes as a bare 500. The SQLite twin can't reproduce it, so the suite stayed green either way.
- Webhook body is capped at 256 KiB (signature verification needs the raw bytes, so the read *must* precede auth) and has a flood limiter.
- `with self._get_connection()` is sqlite3's *transaction* CM — it never closes. All ~20 sites leaked a handle and a WAL read mark; fixed in the helper so no call site changed.
- `_require_admin` re-implemented `api.auth.require_admin`, whose docstring forbids exactly that; three read routes reached past the service into the repo.

**One thing left as-is, deliberately:** purchased Credits still buy nothing — they're in `credit_ledger_entries`, while the metered path spends `user_entitlements.credits`. That's this PR's stated scope, so I only fixed the balance copy, which claimed "Available for model runs and backtests." Wiring them together is a pricing decision, not a bug fix.

`CLAUDE.md`/`.env.example` now record the three env vars and the durability trap: the purchase ledger rides `USERS_DATABASE_URL`, so a deployment that set only `CONTENT_DATABASE_URL`/`AGENT_RUNS_DATABASE_URL` keeps real-money records on ephemeral SQLite.

PR #357, opened 2026-08-15
feat(ui): leaderboard-first landing and home

FlyM1ss:
DO NOT MERGE until Felix has looked at the rendered screens — merging deploys the landing page to prod.

Three asks from Felix:

1. **Leaderboard was ~4 screens down on `/`.** It now sits in the hero's right column — equity chart + top-5 standings, fully above the fold at 1440x900. The agent conversation demo that was there moved down to the Talk act. The `03 — Race` label is dropped; Talk (01) / Test (02) keep theirs, and the Race section keeps the full standings and the rules.
2. **Signed-in users never saw a board.** `index.html` redirects them to `/app`, where Home screen 0 was a second copy of the marketing hero + a "Get Started" button. Screen 0 is now the real Competition Leaderboard (moved off the screen-1 grid, not cloned). The CTA reads by sign-in state.
3. **Plain-trader copy**, to the register adopted in the 2026-08-04 audience-language pass: technology words out, market words kept, icons on the board rules. The Race body now states outright that one of seven AI models finished ahead of the baselines.

## Worth a reviewer's attention

- `js/leaderboard.js` is **untouched** — it cannot be mounted twice (module-level state, ~29 bare-literal ids, chart instance destroyed on every render). Home extends its own existing renderer.
- Two landing guards were rescoped from one file to every component: the disclaimer-freshness check read `Hero.tsx` alone and the allowlisted gloss travelled with the demo; the user-agent check read `Race.tsx` alone and the chart moved. Mutation-tested, 8/8 injected regressions caught.
- Below 1200px the hero stacks and the pager screen is `overflow:hidden`, so the board was **clipped** — rows vanishing with no scrollbar. Fixed; verified 1000x700 → 1920x1080.
- Busters `styles.css` 113→114, `home-page.js` 46→47, checked against `origin/main`. `app.js`/`js/leaderboard.js` unchanged.

Frontend-only: no backend source touched, so no Vercel/Render deploy-skew window. Suite 2879 passed / 78 skipped, unchanged from `main`.


FlyM1ss:
Follow-up filed after merge: #360 — the README hero screenshot (`README.md:29`) depicts the pre-#357 hero (chat demo in the right column, old nav, no ticker) and now advertises a layout the landing page no longer has.

Not introduced by this PR — the shot was already stale from 2026-07-04 — but this change is what made every element of it wrong, so it is recorded here rather than left in a review note.

Issue #365, opened 2026-08-16
leaderboard: config capital ($10k) disagrees with published curves ($100k)


FlyM1ss:
`dashboard/config/leaderboard.json` says `initial_capital: 10000`, but all 12 published contest-window curves in the seed DB were computed at **$100,000**. Nothing reconciles the two, and the cache key can't see the difference.

### Why they diverged

The config changed *after* the board runs were written (2026-07-05 / 07-12):

| Commit | Date | Change |
|---|---|---|
| `0cfc8fb` | 2026-06-18 | added `initial_capital: 100000` |
| `ea1bf2b` | 2026-07-13 | `100000` → `1000` |
| `1dd5816` | 2026-07-20 | `1000` → `10000` ("Scale leaderboard **display** capital to $10k") |

The value is not display-only — `service.py:764` and `:1085` pass it into `strategy_impl.run(...)`.

### What breaks

`_find_cached_run` (`service.py:615`) keys on `(mode, start_date, end_date, llm_model)` — not `initial_equity`. Combined with the `auto_compute` split:

- the **5 baselines** (`auto_compute` unset → true) are recomputed on `?refresh=true` at the config value, **$10k**;
- the **7 LLM entries** (`auto_compute: false`) are only written by `deploy_model_run()`, so they stay at **$100k**.

So one `GET /api/v1/leaderboard?refresh=true` leaves the board comparing $10k baselines against $100k model curves — which is exactly the comparison the board exists to make.

It isn't a wash, because share counts are integers. At the contest window's prices (median DJIA share $249.40, max $910.92):

| | $10,000 | $100,000 |
|---|---:|---:|
| one median share, as % of equity | 2.49% | 0.25% |
| names unbuyable at equal weight | 6 / 30 | 0 / 30 |
| median shares affordable per name | 1 | 13 |
| cash stranded by integer rounding | 34.6% | 4.4% |

`equal_weight_djia` and `buy_hold_djia` would be recomputed with ~35% of capital stranded in cash, depressing them against LLM curves that never were — i.e. the models would look better than they are.

Display is unaffected: `get_leaderboard` reads each row's stored `initial_equity` (`service.py:1205`).

### Severity

Dormant, not live — the daily cron is paused (#352) and `LEADERBOARD_DAILY_AUTO_DEPLOY` is off, so nothing refreshes automatically today. It fires the first time someone forces a refresh.

### Suggested fix

[...]

PR #367, opened 2026-08-16
feat: chart-first landing and screen 0

FlyM1ss:
Puts the equity chart above the fold on both surfaces: `/app` screen 0 gains a real Chart.js chart keyed by the rank rows' swatches, and `/`'s hero becomes 2/3 chart + 1/3 copy with the standings demoted to a one-row chip strip.

**Verified by measurement, not review** — `dashboard/scripts/verify_chart_first_layout.py` (committed) drives a real browser across 8 viewports on both surfaces plus the 3 `/app` no-chart states. All green. Suites: 3083 passed / 84 skipped (backend), 171 / 2 (SDK).

Three plan defects it caught, each invisible to source review:
- The chart reserve was a desktop-only constant — the card ran 77px past the fold at 390×844, where the title, chip and caption wrap and the non-chart height is 335px, not 227px. Now two reserves on a breakpoint-prefixed arbitrary property.
- The script measured during the hero's framer-motion entrance, so every reading was exactly 0.95× (a transformed bounding box). A "height stopped changing" wait doesn't catch it — the scale is *constant* at 0.95 during the delay.
- A children-count wait was satisfied by `app.html`'s static `Loading the standings…` placeholder before any JS ran.

**One correction worth flagging:** the plan justified screen 0's percent axis on issue #365 producing a payload "a dollar axis draws as a 10× scale break". Measured false — `get_leaderboard` rescales every entry by its own stored `initial_equity` and reports one display base, so mixed capital arrives uniform. The percent axis stays, on the reason that does hold: every curve is a ×0.1 rescale of a $100k backtest, so a `$10,749` tick names an account that never existed. #365 is real but costs *returns*, not the axis. Corrected in the plan, the source comment and the test docstring.

Guards strip comments before scanning — mutation-tested, since a class deleted but still named in a comment otherwise satisfies an `in` guard.

Follow-up: `README.md`'s `snapshot.png` goes stale again.


FlyM1ss:
Review pass complete — 15 findings addressed across 8 commits, plus a CodeQL/GHAS sweep. Summary for whoever merges this; the diff is the authority.

## The headline finding

`height: 100%` on the board panel was **inert**. `.home-landing-hero-inner` is `align-items: center`, so the board column was content-sized, the percentage resolved against an indefinite containing block and fell back to `auto`. The panel then overran `.home-landing-hero` (`overflow: hidden`) and was **cut with no scrollbar, no error, no log** — 44–70px depending on viewport.

This PR's own measurement script could not see it: it probed `#homeScreenLanding`, one node *above* the clipping ancestor, and the hero is `height: 100%` of that screen — so it absorbs its own overflow and the screen reads a clean `0` at every viewport. Row-counting was blind for a second reason: what gets cut is the panel's header and footer, not its rows, so the probe recorded `rowsInsideList=7` **and** `rowsInsideHero=7` while clipped. Only panel-box-vs-hero-box, checked at **both** edges, detects it. The script now probes `.home-landing-hero`, and clip is 0px at all 13 viewports.

## Fixes

**`/app` screen 0** — `align-self: stretch` on the board; chart `flex: 0 1 auto` + 132px floor so the illustration yields before the standings; chart teardown on every no-chart path (a failed refresh left a stale chart above fresh rows); `mode: 'nearest'` + a tie filter (every series opens at exactly 0, so `nearest` alone still showed 9 rows / 178px at the leftmost tick — now 1 row / 46.8px); formatted flat x-axis (`Apr 15 … May 12`, was rotated ISO); zero signs `0.00%` to match the rank row beside it; aria-label names the dashed baselines; mock rows carry real `entry_id`s.

**`/` landing** — the chip strip wraps instead of clipping. It was `flex-nowrap` + `overflow-hidden`, measured `scrollWidth 910` against `clientWidth 285` at 390px: four of five chips gone, leaving one name to key five drawn curves, with no fallback legend anywhere on the page. Also YAxis band 44→56 (four of five `$1030` labels were having their `$` sliced), copy column claims the width freed by the 2/3 split, dead `lg:order-*` prefixes removed.

## Two things worth knowing before merging

**1. Seven rows no longer fit at every height — that is the honest outcome, not a regression.** At 1280×720 the panel gets 528px against 642px of content (253 chrome + 187 chart + 202 rows). The list is `overflow-y: auto` precisely so it is the thing that gives; rows below the fold are scrolled, not lost. Visible range is 3/7 to 7/7. The old layout only appeared to show all seven because the overflow was being hidden. Trade-off and the levers are in #370.

**2. Both `/` clamp constants were re-derived by measurement, twice.** Unwrapping the chip strip changed the card's height, so the reserve went 480 → 590 — and that alone was insufficient, because the **300px floor** then became the binding constraint and put the card 95px past the fold at 390×844. Floor lowered to 260. Re-derive both in-browser if the caption, title or strip content changes; the arithmetic is in the comment above them.

## Verification

- Backend suite **3093 passed / 84 skipped**.
- `dashboard/scripts/verify_chart_first_layout.py`: all measurements pass, 13 viewports × 2 surfaces, plus 4 fallback states — including a stale-chart case (success, then a failed refresh) that the fresh-page-per-case harness structurally could not reach before.
- The `/app` source guards now **strip comments before scanning**, mutation-tested in both directions: an implementation commented out but still named in prose fails with the strip and passes without it. That hole is how a previous PR's claim scans went green against the wrong file.
- **CodeQL: 1 alert found** (`py/unused-import`, pre-existing) → fixed → **0 open on `refs/pull/367/merge`**, all three analysis pairs at `results=0`. Confirmed the analysed merge commit carries this branch's head as a parent, rather than trusting the green check — alerts live on the merge ref, and a stale one gives a false all-clear. **Dependabot: 0 open.** Secret scanning returned HTTP 404 with my token, so that one is **unverified, not clean**.

## Follow-ups filed

- #368 — README hero screenshot stale again (carries the verified capture recipe and the do-not-quantise constraint)
- #369 — `/`'s sample chart draws a 7-day axis over a month of returns (pre-existing)
- #370 — screen 0 standings row count, the trade-off above
- #365 — commented with a latent falsy-`or` note on `service.py:1205`, verified unreachable today

PR #374, opened 2026-08-18
fix: log decision-deadline auto-holds

FlyM1ss:
`_maybe_apply_timeout` reattributes a missed-deadline step to
`decision_source="timeout_hold"` with no output at all. A published curve containing
auto-held steps is not the agent's curve, and nothing said it happened.

Emits one line per poll (not per step) from all three sites that drain holds:
`get_current_step`, `get_status`, `drain_expired`. `get_status` matters most — the live
`get_step` path calls it first, which advances `step_index` and makes the later
`step_index == seq` guard fail, so `get_current_step` is never reached on that poll.

The emit helper does not touch `_step_lock` (all three callers already hold it; it is
non-reentrant, and a re-acquire hangs the reaper silently). The step range comes from an
index captured before the loop, since `_advance_step` increments it inside.

T2 of the burst-capacity-safety plan. 9 new tests; full suite 3137 passed / 84 skipped / 0 failed.

### DIFF
--- a/dashboard/backend/domain/backtesting/external_run_service.py
+++ b/dashboard/backend/domain/backtesting/external_run_service.py
@@ -427,11 +427,37 @@ class ExternalBacktestSession:
         self._advance_step(executable=[], decision_source="timeout_hold")
         return True
 
+    def _emit_deadline_hold_notice(self, first_index: int, count: int) -> None:
+        """Print once per poll when this call just auto-held one or more steps.
+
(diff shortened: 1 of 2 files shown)


FlyM1ss:
Merged. One review finding was deliberately left out of scope and is filed as #375: `submit_decisions()` has a fourth deadline branch that bypasses `_maybe_apply_timeout`, so its auto-holds are still silent on the legacy `/api/v1/backtest/*` surface and on the canonical v2 contract. The plan scoped this task to exactly three sites.

PR #380, opened 2026-08-19
feat(ifind): refresh A-share access tokens automatically

MrParamecium:
## Summary

- Add server-side `IFIND_REFRESH_TOKEN` configuration for shared A-share backtests.
- Exchange the long-lived refresh token for a short-lived iFinD access token lazily on the first data request.
- Cache the access token for six days and refresh once after an iFinD 401/403 response.
- Preserve the existing `IFIND_ACCESS_TOKEN` fallback for compatibility.
- Keep iFinD credentials out of frontend code, subprocess arguments, run metadata, logs, and databases.

## Verification

- `3179 passed, 83 skipped`
- `node --check dashboard/frontend/app.js`
- `git diff --check`
- Focused iFinD, A-share integration, router, provider, and frontend tests passed.

## Deployment Notes

- Render `AgenticTrading` has `ENABLE_IFIND_ASHARE=true` and a server-side `IFIND_REFRESH_TOKEN` configured.
- ATL remains a backtest/paper-simulation platform and does not place live orders.
- Alpaca, vn.py, and synthetic-data behavior is unchanged.


FlyM1ss:
Pushed 2 commits to this branch closing review findings — sorry for going straight to the branch rather than a PR into it; revert freely if you'd rather take them differently.

**Correctness**
- The 401/403 `continue` was the only unbounded one in the retry loop, so an expired token seen on the final attempt fell through to `raise AssertionError`. That's outside `IFindClientError`, so `engine.py` doesn't catch it and the backtest dies on a raw traceback. Reserved an attempt for the post-refresh re-send; the terminal raise is now `IFindTransportError` either way.
- `raw_errorcode != 0` rejected a refresh response that omits `errorcode` but carries a valid token — stricter than the data path at `_request_json`. Now presence-gated, matching it.
- Concurrent 401s each forced their own exchange and wiped each other's fresh tokens. Now compare-and-swap under one lock acquisition.
- The six-day cache was per client instance, but a client is built per backtester — so an in-process burst exchanged once per backtest, not once per process as the design specifies. Moved to module scope.
- The exchange had no transport retries; added one, keeping the broad `requests.RequestException` wrap (an escaping one carries the refresh token verbatim).
- All four refresh-failure paths logged nothing, so a revoked token and a renamed upstream field were byte-identical.

**Tests / CodeQL**
- `make_client` left `refresh_token` unset, so it fell back to `os.getenv`. With `IFIND_REFRESH_TOKEN` exported — what this PR asks operators to set up — the suite went **27 failed / 11 passed**. Pinned in the helper and stripped in `conftest.py`.
- Narrowed the threaded test's `except BaseException` (clears CodeQL `py/catch-base-exception`).

Also updated the spec/plan lines that prescribed the two mechanisms above, and documented the `IFIND_*` vars in `CLAUDE.md`. Full backend suite: 3183 passed, 84 skipped.

FlyM1ss:
Follow-up: the `py/catch-base-exception` fix cleared, but my own change then tripped `py/mixed-returns` — the helper I added to log-and-raise refresh failures raised out of its callers' sight, so `_post_token_exchange` looked like it could both return a value and fall off the end. Pushed `f6b5af7` making it build-and-return the error, raised at each call site.

CodeQL on `refs/pull/380/merge` is now **0 results** on both analyses (python and javascript-typescript), and all checks pass.

PR #382, opened 2026-08-19
feat(leaderboard): nof1-style board frame on both charts

FlyM1ss:
Shared visual frame on the Leaderboard tab and Home screen 0, per
`docs/superpowers/specs/2026-08-19-nof1-leaderboard-frame-design.md` §4.

- Reserved right gutter holding an endpoint dot, a dotted stub, the owner name
  and a colour-matched value pill.
- x-axis continues through the gutter to a forward arrowhead.
- The tab's pill now follows `currentChartView` instead of always printing a
  percent beside a dollar axis.

Two things worth knowing before reading the diff:

**The gutter is `layout.padding`, never scale domain.** `resolveHoverTarget`
rejects `x > chartArea.right`, so widening the domain would have made the empty
rail hoverable. Reserving padding keeps `chartArea.right` meaning "end of real
data" and leaves the hover gate untouched.

**§4.1's 3:2 split ships as a ceiling, not as the gutter width.** Rendered, a
fixed 40% reserved 640px on a 1600px canvas to hold ~200px of labels. The gutter
is now `max(measuredBlock, min(width * 0.4, measuredBlock + slack))`: the
measured block is a hard floor so text can never be clipped, and the fraction
caps the rail on narrow canvases. It only binds below `2.5 * floor + 90` px --
about 613px -- so on a 1440 or 1600px tab the slack is what binds and the
fraction is inert; it earns its keep on the home panel and on mobile. Both
arrow-only degradations (too narrow for the widest label, too short to stack
them) are unchanged.

A review pass landed on top of this (`afa1bd6`): the axis arrow was painting
its baseline through descended endpoint labels, the minimum label pitch was
shorter than the pill it has to clear, and the reserved gutter is not the empty
canvas the band note claimed -- the last x tick overhangs into it. Six further
fixes and eight new guards; see the commit body.

Landing hero (PR B) and the backend season block (PR C) follow separately.

https://claude.ai/code/session_01G6tsRGnDnjAm6L1fen4ERi

### DIFF
--- a/dashboard/backend/tests/test_frontend_board_frame.py
+++ b/dashboard/backend/tests/test_frontend_board_frame.py
@@ -442,3 +442,50 @@ console.log(JSON.stringify(out));
     assert second["name"] == "SPY"
     assert second["value"] == "V5"
     assert second["color"] == "#00ff00", "no _style on this dataset -- falls back to borderColor"
+
+
+def test_the_frame_factories_are_explicit_cross_file_exports():
+    """Same contract as `buildEquityCurvesFromEntries`: explicit, not the
+    implicit global these classic scripts share. On rename the implicit form
+    degrades to a chart with no frame, and a frame that silently stops drawing
+    looks exactly like a frame nobody asked for."""
+    assert "window.createEndpointLabelPlugin = createEndpointLabelPlugin;" in _SRC
+    assert "window.createAxisArrowPlugin = createAxisArrowPlugin;" in _SRC
+    assert "window.createEndpointLabelPlugin" in _HOME_SRC
+    assert "window.createAxisArrowPlugin" in _HOME_SRC
+
+
+def test_screen_zero_installs_the_frame_on_its_chart():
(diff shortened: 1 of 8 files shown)


FlyM1ss:
Follow-ups from the whole-branch review, filed rather than carried in this diff:

- #383 — the frontend chart guards never construct a `Chart`; every composition claim is substring-matched. Carries a working headless Chart.js recipe (no jsdom, no node-canvas) and states its adoption cost and its limits.
- #384 — six board-frame guards slice fixed character windows from a call site instead of paren-matching it.
- #385 — screen 0 truncates roster names its own rank list renders in full; measured as safe to change, with the table.

Two review findings were fixed in this branch rather than deferred: duplicated block-width literals (now `BOARD_DOT_GAP`/`BOARD_NAME_GAP`, with a guard that runs the draw hook and diffs its painted width against the measured floor) and a `!= null` endpoint filter that `NaN` passed, which blanked the whole rail.

PR #387, opened 2026-08-20
feat(landing): live board hero

FlyM1ss:
Third of three. #382 (board frame) and #386 (`live` period) are merged; this one puts the **real Competition board on the public landing page**.

Before this, the landing hero and the §Race standings table were invented numbers. Both now render from one `GET /api/v1/leaderboard?period=contest` fetch, shared through a provider so the page fetches once.

**What changed**
- `lib/boardFrame.ts` — the PR A geometry, mirrored for Recharts (the landing is React; it cannot import the vanilla-JS Chart.js plugin). Same constants, same three-pass label stacking.
- `lib/leaderboard.ts` + `useLeaderboard.tsx` — fetch, shape, share.
- `BoardPreview.tsx` / `Race.tsx` — hero chart and standings on live data.
- `EndpointRail.tsx` — the endpoint label rail.
- `storyline.ts` / `Test.tsx` — §02's illustrative run stops naming a real roster model over the real contest window (it contradicted the live hero four screens up: +14.2% vs +0.11%, same model, same window, same base). Placeholders only; one commit to revert if the real model should be named.

**Notes for review**
- The standings table deliberately includes both baselines, unlike `/app`'s models-only rank row — the chart draws those curves, and most models lost to buy-and-hold. Dropping them would make the page more flattering than the truth.
- An empty or model-less 200 now says so rather than rendering a confident board over nothing.
- Bundle rebuilt; `dashboard/frontend/index.html` is hand-patched (#225) so only its two asset-reference lines moved.
- 390×844 does not clear the fold at any reserve — the 260px chart floor binds. The chart itself is above the fold; the chip strip tail is not. Shipped as measured.

https://claude.ai/code/session_01G6tsRGnDnjAm6L1fen4ERi


FlyM1ss:
Review follow-ups filed so they outlive this PR:

- #389 — `GET /api/v1/leaderboard?period=contest` is uncacheable end to end (no `Cache-Control`/`ETag` on the route, `no-store` forced on `/api/*` in `vercel.json`), now served on every anonymous landing view.
- #390 — `service.py:1447` publishes a NULL equity as a real `$0`, which renders as -100% on any base-relative chart. Wire-format fix, needs `/app` moved with it.

Both were deliberately left out of this branch: one changes prod caching behaviour on the highest-traffic anonymous surface, the other changes the payload contract for both frontends. The other 12 review findings are fixed here (`673e702`, `f10fe7d`, `f2fd737`).

PR #391, opened 2026-08-20
fix(home): gate the hint reserve on viewport height

FlyM1ss:
Follow-up to #388, which shipped three defects of mine. Measured under headless Chromium with the board panel **populated the way the app populates it** — seven rows plus the chart wrapper. Omitting the chart is why the earlier pass measured clean.

### 1. The hint reserve was unconditional, and it clipped a control

The board panel is `overflow: hidden` around a chart with a hard `min-height: 132px`, so a fixed 96px hero reserve comes out of the panel:

| viewport height | panel overflow | "See both leaderboards" |
|---|---|---|
| 680 | 12px | reachable |
| 660 | 32px | **unreachable** |
| 640 | 52px | **unreachable** |
| 620 | 72px | **unreachable** |

Clipped silently, since the panel hides its own overflow. It cost rows well above that band too (7 → 5 at 1280×720). A 1366×768 laptop lands inside it once browser chrome and the app's 148px header come off.

Now gated at `min-height: 768px`, measured to be exactly where it costs nothing: at ≥768 the panel keeps all seven rows and overflows by 0; at 760 it starts losing them. Below the gate the hint overlaps the card as it did before any of this — cosmetic, with the footer button and both hero CTAs measured clickable. **Cosmetic overlap beats a clipped control.** Both reserves and the guard now spend one `--home-hint-strip` token; the band is 74px (a 56px hint at `bottom: 18px`), not the "~50px" two comments claimed.

### 2. The ratio ceiling was wrong in the permissive direction

`_MAX_BOARD_GROW` was 1.35; the true ceiling is **1.25**. The binding width is the *widest* screen, not the narrowest — the gap grows to 108.8px while the rail is pinned at 1444, so the copy column bottoms out at 2176+:

| ratio | 1201 | 1600 | 1920 | 2176 |
|---|---|---|---|---|
| 1.25 | +40 | +54 | +13 | **+7.7** |
| 1.30 | +31 | +41 | +0.6 | **−5.0** wraps |
| 1.35 | +22 | +29 | −11.6 | **−17.1** wraps |

The old ceiling admitted two ratios that visibly wrap, having been measured against a table that stopped at 1920. The shipped `1.25` is unchanged and correct; the guard was not.

### 3. Guards that could not fail

Each repair is mutation-verified (13 mutations, all now caught):

- The reserve case **passed with the desktop reserve deleted** — a two-value shorthand made the bottom term a `clamp()`, `_px` raised, and an `except AssertionError: continue` swallowed it, leaving the pre-existing ≤1200px reserve to satisfy it. It now rejects an ungated reserve outright, so defect 1 fails this case.
- The anti-re-anchor assertion read one selector spelling that matched nothing, so it could never fail. It now scans every rule mentioning the hint for `left`/`right`/`inset`/`inset-inline`.
- `_board_block()` took `blocks[0]` — the authoring-order dependency the rail helper was rewritten to remove. Rules are now selected by being **unconditional** rather than first, since the ≤1200px overrides re-declare `flex` and `max-width` too.
- `_declarations` required a trailing `;` and was anchored to start-of-line, so it read nothing from a single-line rule — of which styles.css has many. `_exact_rule_blocks` separately dropped rules preceded by `,`, turning every absence assertion into a pass.

The rail case claimed a jump of 0 "on every pager snap"; equal content rails give equal x only where both caps bind, 1744px up. Below that the screens still differ by 32–47px — pre-existing, and now stated rather than implied away.

`styles.css` cache-buster 117 → 118.

### DIFF
--- a/dashboard/backend/tests/test_frontend_chart_first_home.py
+++ b/dashboard/backend/tests/test_frontend_chart_first_home.py
@@ -864,17 +864,47 @@ def test_the_series_style_helper_is_an_explicit_cross_file_export():
 
 _CSS_COMMENT = re.compile(r"/\*.*?\*/", re.S)
 
-# The hint is `position: absolute; bottom: 18px` and stands ~50px tall, so the
-# strip it occupies is ~68px. The <=1200px branch reserves 96px and documents
-# why ("anything less than ~90px of padding lets the panel slide under it");
-# 90 is that stated floor, asserted rather than the exact 96 so tuning the
-# padding does not have to touch this file.
-_HINT_STRIP_FLOOR_PX = 90
+# `.home-scroll-hint` is absolute at `bottom: 18px` and measures 56px tall, so
+# the band it occupies is 74px. The shipped reserve is larger; this is the floor
+# below which the hint provably sits on top of whatever the hero laid out.
+_HINT_STRIP_FLOOR_PX = 74
 
-# Above this the headline's second line wraps. Measured under headless Chromium
-# with Inter loaded: 1.5 wraps at every width >=1201px, 1.4 wraps at 1920/2560,
-# 1.35 clears 1920 by ~2px. See the rule's own comment for the full table.
-_MAX_BOARD_GROW = 1.35
+# The viewport height at or above which the reserve measured FREE: the populated
+# panel keeps all seven rows and overflows by 0, exactly as it did before the
+# reserve existed. At 760 it starts costing rows, and an UNCONDITIONAL reserve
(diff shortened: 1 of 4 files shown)


FlyM1ss:
Deferred from this PR: #392 (D1), #393 (D6).

Both re-verified at source before filing — #392's claim got **narrowed** in the process: only `test_onboarding_flow_ux.py:165` is provably vacuous (passes with the rule it protects deleted), while the second candidate site does fail its mutation and is merely latent.

PR #395, opened 2026-08-22
feat(leaderboard): add 14 external strategies as baseline classes

TrentonNewWorld:
## Summary
- Adds 14 deterministic (no LLM call) baseline strategies to the leaderboard registry, translated from three external sources backtested in a "Strategy Lab" report: TradingAgents' technical-composite proxy, 6 of QuantConnect's public Investment Strategy Library entries, and 8 of freqtrade-strategies' most prominent strategies (crypto-specific filters identified and stripped rather than blindly ported; GodStra excluded as not portable).
- Adds `_signal_engine.py`, which bridges these daily-scale signals (RSI-14, SMA-200, monthly rebalances, etc.) onto the hourly equity-curve contract `BaselineStrategy.run()` must honor — resamples hourly bars to daily OHLCV, evaluates each strategy's weight function once per trading day using only history strictly before that day (no look-ahead), and marks equity every hour in between.
- Every strategy's lookback adaptively caps at whatever history is actually available rather than requiring a fixed window, since the real leaderboard's contest window (~1 month) is much shorter than the full year these strategies were originally validated against.
- Registers all 14 in `registry.py` and adds matching entries to `leaderboard.json`'s `strategies` array (a registered class alone is inert — it only becomes visible/live once it also has a config entry, since that's the only call site for the registry).
- Also includes two earlier, unrelated fixes made in the same working session: the Windows venv-path + repo-root-relative `DATABASE_PATH` bugs (`fix(backtests)`), and a risk-gated Alpaca live-trading broker + a new Mission Control wallet/holdings page (`feat(trading)`).
- Along the way, fixed two pre-existing gaps caught while verifying this: `mission_control.py` was leaking raw exception text via `"error": str(e)` (violates `test_error_detail_sanitization.py`'s CodeQL-motivated guard), and `test_app_composition.py`'s route contract didn't yet include the two Mission Control routes.

## Test plan
- [x] `pytest dashboard/backend/tests/domain/leaderboard/test_external_strategies.py -v` — 58 new tests (registry identity, key resolution, required_symbols, run() smoke tests against synthetic hourly bars, including a short-history graceful-degradation case).
- [x] Full backend suite (`pytest dashboard/backend/tests/`) — only the same 8 pre-existing, unrelated failures remain (iFinD, vnpy, optional-SDK, `test_engine_move`, `test_canonical_consumers`, `test_router_move::test_no_circular_imports`) plus one pre-existing CSRF test error; confirmed via a clean `git stash` baseline that none are caused by this branch.
- [x] Verified end-to-end against real Alpaca market data on a local dev server: all 26 leaderboard entries (12 original + 14 new) return sane, finite equity curves via `GET /api/v1/leaderboard?period=contest`, and all 14 new strategies appear by name in the frontend's chart legend/picker on the Competition tab.


Allan-Feng:
Thanks for putting this together. We really appreciate the work that went into the strategies.

A suggestion: rather than adding all of them to the leaderboard right away, it would be great to pick 2–3 of them and run them through the ATL backtest workflow first, and compare their results with the existing market indices/ buy-and-hold baselines.

After testing the strategies, Agent Supermarket would be a great place for them. That’s where we’d love community agents to land, so people can try them, clone them, and iterate.

Happy to talk through details anytime. Feel free to ping me (allanfeng) on our Discord: https://discord.gg/9HnQ6XDG98

FlyM1ss:
Review pass done; fixes pushed as three commits on this branch (`b154da8d`, `32e2c14e`, `998ef916`). Full backend suite green locally (3421 passed), CodeQL alert #1282 (`py/unused-import`) closed.

**Strategies (`b154da8d`)** — behaviour: turn_of_month bought SPY on the last day of every contest (last date flagged as month-end) and a zero open started a phantom hold; capm_alpha_ranking hardcoded 0.5 weight (half the book in cash with one qualifier) and its OLS variance was over the full window while the covariance used the symbol's valid days; `pct_change()` padded gaps into 0% days (capm, volatility_effect); bandtastic could never trade (50-day floor vs ~44-day window); trendrider's SMA-200 went permanently NaN after one gap day; rsi NaN with no down moves, adx NaN on a flat stretch; `build_price_cache` dropped any symbol missing the first timestamp for the whole run. Structure: hlhb/trendrider/supertrend_triple precompute indicators once per symbol (causal → bit-identical, pinned by `test_indicators.py`); `daily_history` maps dates once not 5×; dead `lot_size` removed; the 12 `required_symbols` + 14 `num_trades` copies moved into `base.py`. Tests: the multi-symbol smoke tests were vacuous (fixture SYM0..29 vs DJIA-30 default → `run()` returned `[]` before touching anything) — they now pass symbols and assert curve length; added a reference-window engine test.

**Live broker / Mission Control (`32e2c14e`)** — `/api/v1/mission-control/overview` is now admin-gated (it exposes the real account and whether execution is armed); live snapshot errors no longer 500 the page; 30 s cache; page sends credentials + escapes output. `alpaca_live_service` runs the decision through the LLM validator before building orders; feed resolution errors are fatal instead of swallowed; `ALPACA_LIVE_*` stripped in conftest; `alpaca_live.json.example` + CLAUDE.md docs.

**Backend (`998ef916`)** — shared `resolve_python_exe`/`resolve_env_path` replace four copies of the Windows-venv / relative-`DATABASE_PATH` logic; leaderboard service builds `_cached_run_index` once per request (was `_find_cached_run` × 19 entries × 3 call sites) with a single-flight lock per window; `leaderboard.js` styles preset-less baseline entries as `strategy`, not `team`.

**Not changed, flagged for follow-up**
- Empty weights `{}` = liquidate (sanctioned by the `_signal_engine` docstring); entry/exit strategies keep `held` instead — worth a deliberate decision.
- trendrider's fixed `ema(close, 50)` is all-NaN below 50 rows, so the golden-cross entry / bearish-cross exit never fire on the leaderboard window (the other four signals still do). Same cap as bandtastic would fix it but breaks the precompute.
- `TEAM_COLOR_PALETTE` has 10 colours for 19+ entries → collisions; presets for the new labels would need to avoid `MODEL_COLOR_PALETTE`.
- Truthy-env parsing is duplicated across the two live services and the leaderboard flags — a shared helper.
- User-facing docs now stale (not edited here): `docs/source/lab/live_trading.rst` intro, `docs/source/lab/getting_started.rst` (~L130–154), README "Go live" bullet.

FlyM1ss:
Overall looks good. I also recommend to test 2 to 3 strategies first.

FlyM1ss:
**Next step: re-scope to 3 strategies as Agent Supermarket templates** (per Allan's comment above)

Why not just trim the list: the 14 strategies here are Python leaderboard baselines, but the ATL backtest workflow and the Supermarket only run prompt/pipeline agents (`dashboard/config/marketplace.json`; precedent: **Blue-Chip Steady** / **Even-Split Dow** are prompt renditions of leaderboard baselines). "Test 2–3 first" therefore means rewriting them as prompt templates, not shipping a subset of the Python.

Shortlist — the three whose rules are expressible from what the agent actually sees each hour (`price, rsi, macd, macd_signal, sma20, sma50, bb_upper, bb_lower`; no price history):

1. `universal_macd` — MACD > 0 (identical to its EMA12/EMA26 zero-cross)
2. `tradingagents_composite` — RSI + MACD + SMA composite
3. `bandtastic` — Bollinger bands + SMA20/SMA50 (slow EMA → SMA50)

The other 11 need ADX/ATR/SMA-200 or multi-week return history the agent never receives — park them.

Actions (@TrentonNewWorld; a maintainer picks this up if there is no reply within a week):

1. Add the three as `marketplace.json` templates (`category: us_stocks`, one `simple_instruction` step each; crossover rules become level rules).
2. Backtest each from the dashboard Backtest page on the contest window (2026-04-15 → 2026-05-15) and post return / Sharpe / max drawdown next to the DJIA and buy-and-hold lines the chart already draws. Merge is gated on those numbers.
3. Remove the 14 registry + `leaderboard.json` entries from this PR; move the Python strategy modules and their tests to a separate draft PR as reference.
4. Split the unrelated commits into their own PRs so they stop being blocked here: `51f38be8` Windows venv / `DATABASE_PATH` fix (mergeable now) and `af74a89d` Alpaca live broker + Mission Control.

PR #448, opened 2026-09-10
fix(backtest): escalate the output ceiling on the first retry, not the fifth

Anish-1352:
## Problem

ATL ships `reasoning.max_tokens = 2048` against `max_tokens = 2000`. Reasoning
tokens are output tokens, so a model that uses its budget is truncated inside
the thinking block and returns a reasoning trace with no text. Every observed
failure came back at exactly the ceiling; none below it.

An empty reply is what a reply looks like when reasoning ran out the output
ceiling before any text was emitted. Re-asking at the ceiling that just failed
re-issues the same request under the same constraint that proved insufficient.
It sometimes succeeds, when the model happens to reason less. The single-prompt
loop does that four times before raising it.

`pipeline_runner` already disagrees with itself here: an empty pipeline
response is retried ONCE, immediately at `RECOVERY_MAX_OUTPUT_TOKENS`. The
same failure costs one extra call on that path and five on this one. This
makes the two agree.

## What changes

The first request is deliberately untouched, and untouched in call shape, not
merely in resolved value: `max_tokens` is passed only when it is actually
being raised, so an unescalated attempt is the identical call it is today. A
decision that succeeds first time — the overwhelming majority — issues exactly
one request, unchanged. That is what keeps this a cost change rather than a
behaviour change, and it is why the existing
`test_portfolio_final_no_text_retry_preserves_reasoning_and_increases_budget`
still passes unmodified.

Also sets `recovery_spent` when the escalated retry is what produced the text.
Without it the post-parse truncation retry would re-issue an identical request
at the same ceiling, which upstream's own comment says has nothing different
left to ask for.

Behind `LLM_ESCALATE_CEILING_ON_RETRY`, default off — nothing changes until
enabled. Happy to make it the default if preferred; the change only affects
requests that had already failed.

## Measured

Three interleaved pairs, 42 decisions per side. Interleaved because the same
configuration on the same window gave 2.21, 2.43 and 3.36 attempts/decision
across runs — a single A/B here measures the day, not the change.

| | off | on | delta |
|---|---|---|---|
| attempts / decision | 2.67 | 1.88 | −29.5% |
| retries | 70 | 37 | −47.1% |
| decisions needing ≥4 attempts | 14/42 | **0/42** | Fisher p < 0.001 |
| input tokens | 224,255 | 150,130 | −33.1% |
| cost | $0.0596 | $0.0455 | −23.6% |

No decision exceeded three attempts with the flag on. Variance collapses too:
off ranged 2.21–3.36 across runs, on gave 1.93 / 1.93 / 1.79.

**The tail is the point.** A decision that exhausts the retry loop falls back
to the rule-based agent, and nothing is persisted to record that it happened.
Fourteen such decisions became zero.

Quality is unchanged within noise (returns −0.0052 vs −0.0017 mean over three
runs a side, about one standard deviation). Three runs cannot show it got
better, but it did not get worse — unlike raising the default ceiling, which
was adverse on the one window measured (−0.19% → −0.71%).

## Scope

This failure does not occur on the multi-step pipeline path: 168/168 calls
parsed with zero ceiling hits across three market regimes, against 71% ceiling
hits on the single-prompt path. Decomposing agents into pipeline steps would
avoid it entirely.

The leaderboard runs the single-prompt path — the seed leaderboard runs
measured 1.000 calls per bar (nemotron 0.994) — so this fix applies where the
platform actually operates. Decomposition is a larger change: calls per
decision equals the configured step count exactly (measured 3.000 and 5.000),
and each step's prompt embeds all prior outputs, so input compounds. Whether
that nets out cheaper than retries is unmeasured.

## Tests

4,222 passed, 160 skipped, 0 failed. Default-off equivalence asserted by its
own test.


FlyM1ss (approved):
Reviewed. All six findings from the review pass are addressed in `e006d6b` (pushed to this branch with `maintainer_can_modify`), CI and both CodeQL analyses are green.

Note for transparency: I pushed the fix commit here, so this approval covers the review findings and their fixes, not an independent read of my own changes. Details in the [summary comment](https://github.com/Open-Finance-Lab/AgenticTrading/pull/448#issuecomment-5622570975).

PR #490, opened 2026-09-18
feat: draw the universal top nav on /admin

FlyM1ss:
Replaces `/admin`'s bespoke mock header with a retyped copy of `app.html`'s top bar, and adds the guard that keeps it a copy.

Chunk C1 (§0) of `docs/superpowers/plans/2026-09-18-admin-console-port.md`. Answers the second of the two user reports the plan carries.

## What
- `admin.html` — the `.header` block + three sprite symbols. Six deltas, all forced by app.js not loading here; enumerated in the markup comment and normalised by the guard.
- `admin.css` — `.header` family + seven breakpoints retyped from `styles.css`, on a `:root` alias block carrying styles.css's **values** (not this file's nearest equivalent — substituted tokens are how two copies of one header become two headers). Bare `header{}`, `.brand`, `.product-nav` deleted; `.workspace`'s `80px` literal becomes `--header-height`.
- `admin-shell.js` — account element IDs renamed to app's; app.js's `#navMenuToggle` handler copied (below 900px the hamburger is the only route to the nav).
- `test_admin_header_parity.py` (new) — extracts both `<header>` blocks, normalises the six deltas, asserts equality. Each substitution must prove it fired.
- `test_admin_page_shell.py` — the "no styles.css token names in admin.css" guard becomes "every token used here is declared here". The old form banned exactly what this port needs; the new one checks the property it was protecting, across every token.

Design N7 ("no Admin menu item on /admin") is overridden — see the commit body.

## Test plan
- [x] `pytest dashboard/backend/tests/ -q` → 4920 passed, 162 skipped (baseline 4912)
- [x] Parity guard discriminates: flipping one word in admin.html's header fails `test_the_two_headers_are_the_same_header`
- [ ] Cache-busters: `admin.css?v=4`, `admin-shell.js?v=5`, all pins re-grepped


FlyM1ss:
Deferred from this PR: `docs/superpowers/plans/2026-09-18-admin-console-port-deferred.md` §0.C1 (D2) — no drift guard on the ported CSS declarations; the parity test compares markup only.

PR #491, opened 2026-09-18
feat: port Account Management and Activity to /admin

FlyM1ss:
DO NOT MERGE until #492 is resolved — that PR (different author) ports the same two views onto `/admin` under different element IDs (`#account`/`#activity` vs this PR's `#accountManagementView`/`#activityView`). Merging both ships two copies of the same markup. A human must pick one.

## Summary

Chunk C2 of `docs/superpowers/plans/2026-09-18-admin-console-port.md`.

- `app.html:2128-2284` (`#adminView`) and `:2286-2305` (`#adminGrantReasonDialog`) move onto `admin.html` as `#accountManagementView` + `#activityView`, element IDs unchanged so `admin-credits.js` binds without a rename.
- Shell gains `account-management` / `activity` routes. The non-Analytics entries are now one named `CONSOLE_ROUTES` list, read both by `ROUTES` and by the `pageControls`/`freshnessLegend` suppression — a route added to one and not the other used to be possible.
- `admin.css` gets the `.admin-credits-*` family retyped from `styles.css` (`/admin` never loads `styles.css`), on styles.css's own token names via the existing `:root` alias block.
- Sprite gains `#icon-wallet`, `#icon-search`.
- `admin.css?v=5`, `admin-shell.js?v=6`, pins reconciled across tests.

## Deliberately left behind

Four parts of `#adminView` did not come across: its page-header/error/success/refresh chrome, `#adminTabs`, `#adminStats` (`admin-live.js` already renders the same `/api/admin/stats` counters on Overview), `#adminLegacyUsersPanel` (dead — `hidden` in `app.html` with nothing anywhere unhiding it). `#creditsRefundDialog` stays on `app.html`; it belongs to Credits & Billing. Their CSS was not ported either.

## The rail hrefs do NOT flip in this PR

`admin-credits.js` is rewired onto `AdminShell` in C3. Until it is, a rail entry pointing at `#account-management` would open a permanently empty table — worse than the cross-surface jump it replaces — and merging here auto-deploys prod. The three links (both rail entries plus `#evidenceAccount`) flip in C3; `6bc00360` records that on `main`.

## Test plan

- [x] `pytest dashboard/backend/tests/ -q` → 4925 passed, 162 skipped (baseline 4920)
- [x] New: required-ID set is **derived from `admin-credits.js`** by regex, not retyped — nothing loads that module on `/admin` yet, so a node dropped in the move has no runtime symptom and this is the only thing that can see it
- [x] New: both views ship `hidden`; suppression asserted against `CONSOLE_ROUTES`, not a per-route literal
- [x] New: dropped panels and their CSS assert-absent, so a later chunk cannot re-add a second owner
- [x] `#67e8f9` guard narrowed from a file-wide ban to count+declaration — the one occurrence is verbatim from `styles.css:13361`


FlyM1ss:
Superseded by #492, which landed the same Account Management / Activity port. This branch conflicts with `main` in 5 of its 6 files, so there is nothing left here to rebase onto.

The review of this PR did surface three defects that are real on `main` today (they shipped with #492): the credits status line is never retired when the shell routes away, `setStatus` drops a class the markup ships, and the Grant Pool region lost its accessible name in the `<section>`→`<div>` demotion. Those are fixed in #494.

Closing unmerged; leaving the branch in place.

PR #494, opened 2026-09-18
fix: scope the admin credits status line to its own routes

FlyM1ss:
Three defects on `/admin`, found reviewing #491 and re-verified against `main` (all three shipped via #492; #491 itself is superseded).

**1. The credits status line is never retired.** `#adminCreditsStatus` sits outside `#accountView` and `#activityView` because both write to it, so no view hides it the way `/app`'s `display:none #adminView` did. `route()` never clears it, so "Grant Pool funded." stays at the top of `<main>` on Overview and Providers, with `aria-live` announcing it as current. `admin-credits.js` now clears it on the route event it already listens to.

**2. `setStatus` dropped a class the markup ships.** It assigned `className` outright as `credits-status`, but the node ships `credits-status admin-credits-status`, so the first write lost `.admin-credits-status`'s margin — on `/admin` and `/app` both.

**3. The Grant Pool region lost its accessible name.** The `<section>`→`<div>` demotion that stopped a nested `</section>` truncating `PANEL_REGION` was right, but `aria-labelledby` is inert without a role. `role="group"` keeps the truncation fix and the name.

Tests: 3 new, each watched failing first. The markup guard is case-insensitive and has its own companion test, for the reason the `SCRIPT_TAG` guard beside it documents — a check that proves a negative is only as good as the spellings it can see. Full suite: 4931 passed, 162 skipped.

Cache-busters: `admin-credits.js` was pinned at two versions by two documents — `app.html` at `v=6` (4 pins) and `admin.html` at `v=2` (2 pins). Both bumped.

https://claude.ai/code/session_01NUSgbaFPuckwMHnTgLmFw6


FlyM1ss:
CodeQL clean on the merge ref: the analysis is diff-informed (`Persisted 11 diff range(s) across 8 file(s)`), so `results=0` on both `python` and `javascript-typescript` means zero alerts on the code this PR changes — not an unpopulated query.

Three pre-existing `py/unused-import` alerts in `test_admin_absorption.py` are also handled: #1347 (`re`) and #1348 (`pytest`) deleted, #1349 (`run_node`) closed by *use* — the new status-line tests call it. They are tracked against `main`'s ref, so they close on the next default-branch scan after this merges.

Follow-up filed as #495 (two dead `.admin-workspace-content` rules in `admin.css`), found while correcting a comment here. No overlap with this diff.

